PR #10, opened 2026-09-16
feat(config): add Config type, loader, and startup validation

### LINKED ISSUE #1: Configuration schema, defaults, and startup validation
## What to build

A `Config` type and loader that reads a YAML file (path from a flag or env var) into a global section (listen address, served config path, poll interval, per-request timeout, max response size, edge entrypoint names, log level) plus a list of downstreams (name, API address, traffic address, allowed downstream entrypoints, optional auth, optional TLS, optional priority offset, optional staleness limit). Every field except downstream name and addresses has a sensible default — no duration field defaults to an empty string. Unknown fields are rejected, durations parsed, downstream names checked for uniqueness and safe characters. A validation failure exits non-zero with a message naming the offending field.

Out of scope for this ticket: polling, HTTP serving, transformation — those come later. This ticket is done when a config file can be loaded and validated, nothing more.

## Acceptance criteria

- [ ] A minimal valid config (just downstream names/addresses) loads successfully with every other field defaulted
- [ ] Missing downstream address, duplicate downstream name, unparseable duration, and unknown field each produce a startup failure naming the specific bad field
- [ ] No duration field silently defaults to empty string
- [ ] Tests drive the loader against real YAML fixtures (good and bad), not by inspecting internal struct state

gringolito:
## Summary

Adds `internal/config` with a `Config` type, a `Load(path string) (*Config, error)` function, and 7 tests that drive the loader against real YAML fixtures.

The loader reads a YAML file, applies defaults for every optional field, and validates before returning. A failure exits with an error that names the specific bad field.

## Why

This is the foundation every subsequent ticket depends on. Polling, HTTP serving, and transformation all need a loaded, validated config to work from.

Notable choices:

- Duration fields are `string` in the internal `rawConfig` struct and parsed manually in `convert()`. This is the only way to get error messages that name the field (`poll_interval: invalid duration "bad"`) instead of the Go type.
- `yaml.Decoder.KnownFields(true)` rejects unknown YAML keys. The error from yaml.v3 includes the line number and field name.
- `Auth` and `TLS` stub types are present on `Downstream` now so configs with those optional blocks aren't rejected by `KnownFields`. Their fields will grow in later tickets.

Closes #1


internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	PriorityOffset     int      `yaml:"priority_offset"`
+	StalenessLimit     string   `yaml:"staleness_limit"`
+}
+
+func rawDefaults() *rawConfig {
+	return &rawConfig{
+		Listen:          DefaultListen,
+		ConfigPath:      DefaultConfigPath,
+		LogLevel:        DefaultLogLevel,
+		PollInterval:    "30s",
```
gringolito: **Standards — Duplicated Code.** `DefaultPollInterval` (line 18) and `DefaultRequestTimeout` (line 19) are declared and never referenced anywhere; these literals are the actual defaults. Go allows unused constants, so `go build` and `go vet` stay silent while the constant and the literal drift apart. Either set these from the constants (`PollInterval: DefaultPollInterval.String()`) or delete the two constants. Whichever way, they should have one source of truth.

Also worth knowing: the test asserts `PollInterval > 0`, not that it equals the default, so a silent default change would not fail the suite.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	CA                 string `yaml:"ca"`
+	Cert               string `yaml:"cert"`
+	Key                string `yaml:"key"`
+	InsecureSkipVerify bool   `yaml:"insecure_skip_verify"`
+}
+
+// rawConfig mirrors Config with duration fields as plain strings so that
+// yaml.v3 decodes them without interpretation; we parse them ourselves to
+// produce error messages that name the offending field.
+type rawConfig struct {
```
gringolito, line 65: **Standards — Duplicated Code (judgement call, no action recommended).** This mirror struct duplicates the shape of `Config`. It earns its keep right now: it buys strict `KnownFields` decoding, duration errors that name the field, and defaults pre-merge for ~30 lines. A custom `UnmarshalYAML` on `Config` would remove the mirror but add per-field boilerplate and subtler decode edge cases. Revisit only when a third duration-typed field group appears; acting today is churn.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	Downstreams     []Downstream
+}
+
+// Downstream describes one upstream Traefik instance.
+type Downstream struct {
+	Name               string
+	APIAddress         string
+	TrafficAddress     string
+	AllowedEntrypoints []string
+	Auth               *Auth
```
gringolito, line 41: **Standards — Speculative Generality (judgement call).** `Auth`, `TLS`, `PriorityOffset`, `AllowedEntrypoints` and `EdgeEntrypoints` (line 30) are declared but never defaulted, validated, or exercised by a single fixture. If nothing consumes them yet and the next tickets are not imminent, trim to what the spec needs. The counter-argument is load-bearing though: `KnownFields(true)` rejects `auth:`/`tls:` blocks unless the fields exist, so removing them would make configs the spec calls valid fail to parse. Suggest leaving them and adding a fixture that exercises each one so they are not untested declarations.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	d, err := time.ParseDuration(s)
+	if err != nil {
+		return 0, fmt.Errorf("%s: invalid duration %q: %v", field, s, err)
+	}
+	return d, nil
+}
+
+// parseOptionalDuration allows an empty string (meaning "not set") and returns
+// zero in that case, which callers interpret as "no limit".
+func parseOptionalDuration(field, s string) (time.Duration, error) {
```
gringolito, line 180: **Standards — Data Clumps (judgement call, no action).** `(field, s)` travels together through `parseDuration` and `parseOptionalDuration`. A `durationField` type would be over-engineering at this size. Flagging for the record only.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Config holds the complete, validated application configuration.
+type Config struct {
+	Listen          string        `yaml:"listen"`
+	ConfigPath      string        `yaml:"config_path"`
+	PollInterval    time.Duration
+	RequestTimeout  time.Duration
+	MaxResponseSize int64
+	EdgeEntrypoints []string
+	LogLevel        string
```
gringolito, line 31: **Standards — Primitive Obsession (judgement call, no action).** `LogLevel string` holds a constrained value in a bare string. Not worth a dedicated type until something validates it. Skip for now.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	APIAddress         string   `yaml:"api_address"`
+	TrafficAddress     string   `yaml:"traffic_address"`
+	AllowedEntrypoints []string `yaml:"allowed_entrypoints"`
+	Auth               *Auth    `yaml:"auth"`
+	TLS                *TLS     `yaml:"tls"`
+	PriorityOffset     int      `yaml:"priority_offset"`
+	StalenessLimit     string   `yaml:"staleness_limit"`
+}
+
+func rawDefaults() *rawConfig {
```
gringolito, line 87: **Standards — Mysterious Name, mild (judgement call, no action).** `rawDefaults` exists mainly so `Load` can decode into a pre-populated struct. Inlining the literal into `Load` removes one indirection. Negligible either way.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+		ConfigPath:      DefaultConfigPath,
+		LogLevel:        DefaultLogLevel,
+		PollInterval:    "30s",
+		RequestTimeout:  "5s",
+		MaxResponseSize: DefaultMaxResponseSize,
+	}
+}
+
+// Load reads the YAML file at path, applies defaults, and returns a validated Config.
+func Load(path string) (*Config, error) {
```
gringolito, line 100: **Spec — partial.** Spec: *"a loader that reads a YAML file (path from a flag or env var)"*. No flag parsing and no environment lookup exists anywhere in this diff (`flag.`/`Getenv` across `*.go` is zero hits); `path` arrives purely as a parameter. The resolution mechanism is the missing half of that requirement. Reasonable to defer if a `main` package lands in the next ticket, but the ticket should say so rather than leaving it implicit. Note there is no test that pins which env var or flag name is used, so whenever it lands it will be unpinned behaviour.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	cfg, err := convert(raw)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := validate(cfg); err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
```
gringolito, line 124: **Spec — partial.** Spec: *"A validation failure exits non-zero with a message naming the offending field"*. The error-message half is fully implemented and well done (`validate` names the field and position for every case). The exit-non-zero half has no implementation at all: no `main`, no `os.Exit`, so nothing in this diff can exit non-zero. Correct to defer the exit path until there is a binary, but that makes this acceptance criterion only half-met as it stands.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+// Config holds the complete, validated application configuration.
+type Config struct {
+	Listen          string        `yaml:"listen"`
+	ConfigPath      string        `yaml:"config_path"`
+	PollInterval    time.Duration
+	RequestTimeout  time.Duration
+	MaxResponseSize int64
+	EdgeEntrypoints []string
```
gringolito, line 30: **Spec — judgement call.** Spec: *"Every field except downstream name and addresses has a sensible default"* and *"edge entrypoint names"*. This is the one global field with no explicit default in `rawDefaults()`; it stays nil. An empty list is defensible as the sensible default for a list, but note it is inconsistent with the other six global fields, which all get an explicit value. If the downstream default is meant to mirror an upstream Traefik's usual `web`/`websecure`, that intent is invisible here.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+var validName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+const (
+	DefaultListen          = ":8080"
+	DefaultConfigPath      = "/config"
+	DefaultLogLevel        = "info"
+	DefaultPollInterval    = 30 * time.Second
```
gringolito, line 18: **Spec — judgement call (implemented correctly).** Spec: *"no duration field defaults to an empty string"*. `DefaultPollInterval`/`DefaultRequestTimeout` are typed `time.Duration` here, but the defaults that actually reach the raw struct are the strings `"30s"`/`"5s"` at lines 92-93. Behaviour is right: an unset duration is never an empty string in the returned `Config`, and a bad one fails with the field named. Flagging only because the typed constants sitting beside the string defaults invites the reader to assume one feeds the other.

internal/config/config.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, err
+	}
+	requestTimeout, err := parseDuration("request_timeout", r.RequestTimeout)
+	if err != nil {
+		return nil, err
+	}
+
+	downstreams := make([]Downstream, len(r.Downstreams))
+	for i, rd := range r.Downstreams {
+		sl, err := parseOptionalDuration(fmt.Sprintf("downstream[%d].staleness_limit", i), rd.StalenessLimit)
```
gringolito, line 142: **Spec — judgement call (implemented correctly).** Spec: *"optional staleness limit"*. Unset becomes the zero `time.Duration` through `parseOptionalDuration`, and the criterion that no *duration* field defaults to an empty string still holds because the value is a typed duration, not a string. Good. Worth a one-line comment at the call site later stating that zero means "no limit", so the next ticket does not have to read `parseOptionalDuration`'s doc comment to learn it.

gringolito (commented):
## Two-axis review (Standards / Spec)

**Spec source.** The PR body says `Closes #1`, so the spec is issue #1 *Configuration schema, defaults, and startup validation*.

Inline comments carry the line-level findings. Both sub-agents' line numbers were re-verified against the file before posting.

### Standards

No hard violations are citable because the repo documents no standards. Five judgement calls from the smell baseline, four of which recommend no action at this size:

- **Duplicated Code** — `rawConfig`/`rawDownstream` mirror `Config`/`Downstream`. Earns its keep: strict `KnownFields` decoding, named-field duration errors, and defaults pre-merge in ~30 lines. A custom `UnmarshalYAML` per duration field would remove the mirror but add boilerplate and subtler decode edges. Revisit only if a third duration-typed group appears.
- **Data Clumps** — `(field, s)` travels through `parseDuration`/`parseOptionalDuration`. Harmless at this size.
- **Primitive Obsession** — `LogLevel string` for a constrained value. Not worth a type until log-level validation exists.
- **Speculative Generality** — `Auth`, `TLS`, `PriorityOffset`, `AllowedEntrypoints`, `EdgeEntrypoints` are declared but never defaulted, validated, or exercised by any fixture.
- **Mysterious Name (mild)** — `rawDefaults()` exists mainly so `Load` decodes into a populated struct. Negligible.

One finding surfaced during aggregation, not in either axis report: `DefaultPollInterval` and `DefaultRequestTimeout` are declared and never referenced, while `rawDefaults()` hardcodes `"30s"`/`"5s"`. Go allows unused constants, so nothing catches the drift.

Positive: `validate` is a clean named-error cascade, the yaml-tag split between raw and public types is consistent, and `parseOptionalDuration` shares `parseDuration` rather than duplicating the error format.

### Spec

No scope creep; every hunk maps to a spec item. Two partials, three judgement calls. All four acceptance criteria pass.

- **partial** — *"reads a YAML file (path from a flag or env var)"*. Nothing reads a flag or an environment variable; the path is only a parameter. The wiring half of the requirement is unaddressed, though `main` doesn't exist yet, so only the resolution mechanism is genuinely missing.
- **partial** — *"A validation failure exits non-zero with a message naming the offending field"*. Errors name the field, but there is no process exit path at all (no `main`, no `os.Exit`), so the exit-half is unimplemented.
- **judgement call** — *"Every field except downstream name and addresses has a sensible default"*: `edge_entrypoints` gets no explicit default, unlike every other global field.
- **judgement call** — duration defaults are raw strings re-parsed via `parseDuration`. Correct, slightly indirect, sound.
- **judgement call** — unset `staleness_limit` becomes a `time.Duration` zero via `parseOptionalDuration`. Fine: explicitly optional, and the zero is not an empty string.

Field-default audit (spec list vs `rawDefaults()`): listen `:8080` ✓, config_path `/config` ✓, poll_interval `30s` ✓, request_timeout `5s` ✓, max_response_size 10 MiB ✓, log_level `info` ✓, edge_entrypoints nil ✗. Downstream `allowed_entrypoints`/`auth`/`tls`/`priority_offset`/`staleness_limit` are zero-valued, which is correct because the spec marks them optional; `name`/addresses are required and validated.

| Criterion | Verdict | Evidence |
|---|---|---|
| Minimal config loads with defaults | PASS | `TestLoad_MinimalValid`; `minimal_valid.yaml` carries only name+addresses |
| Missing address / duplicate name / bad duration / unknown field fail naming field | PASS | Four tests pass; errors name `api_address`, `traffic_address`, the duplicate value, `poll_interval`, `typo_field` |
| No duration defaults to empty string | PASS | Durations are `time.Duration` in public `Config`; `""` exists only in the raw layer |
| Tests use real YAML fixtures, not internal struct state | PASS | All 7 tests call `config.Load("testdata/*.yaml")` and inspect public `Config` only |

Build/test evidence: `go build ./...` clean; `go test ./internal/config/ -v` 7/7 PASS.

---

**Standards: 6 findings, 0 hard violations; worst is the dead `DefaultPollInterval`/`DefaultRequestTimeout` pair plus the five unexplained struct fields.** **Spec: 5 findings, 4 criteria PASS; worst is the partial gap on flag/env-var path resolution.**

internal/config/config.go
gringolito: Fixed in c64d58f. rawDefaults now calls DefaultPollInterval.String() and DefaultRequestTimeout.String(), so the typed constants are the single source of truth. A silent default change will change the constant, the string passed to parseDuration, the parsed time.Duration, and the > 0 assertion in the test all together.

internal/config/config.go
gringolito: Agreed, no action. The mirror stays. Will revisit if a third duration-typed group appears.

internal/config/config.go
gringolito: Fixed in c64d58f. Added TestLoad_DownstreamOptionalFields with testdata/downstream_optional_fields.yaml. The fixture sets auth (username+password), tls (ca+insecure_skip_verify), allowed_entrypoints, priority_offset, and staleness_limit; the test asserts each field round-trips. All five fields are now exercised by at least one passing test.

internal/config/config.go
gringolito: Noted. Leaving as-is.

internal/config/config.go
gringolito: Noted. Leaving as-is until log-level validation is added.

internal/config/config.go
gringolito: Noted. Leaving rawDefaults as a named function; the pre-populated struct pattern is clearer than an inline literal inside Load.

internal/config/config.go
gringolito: Deferred to the main package ticket. Load deliberately accepts a path parameter so the library stays testable without flag parsing. The wiring (which env var name, which flag name, precedence order) belongs in main where both are visible at once. That ticket should include a test that pins the env var name and the flag name so they cannot drift.

internal/config/config.go
gringolito: Deferred. There is no binary yet. The exit path belongs in main alongside the flag/env wiring. The error message half is complete; os.Exit lands when the main package is added.

internal/config/config.go
gringolito: Fixed in c64d58f. rawDefaults now sets EdgeEntrypoints: []string{}, consistent with the other six global fields. Left as an empty list rather than a Traefik-specific default (web/websecure) because the correct set depends on each deployment's Traefik config.

internal/config/config.go
gringolito: Fixed in c64d58f as part of the dead-constants fix. rawDefaults now calls DefaultPollInterval.String() and DefaultRequestTimeout.String(), so the typed constants directly produce the strings that reach the raw struct. No reader needs to wonder whether one feeds the other.

internal/config/config.go
gringolito: Fixed in c64d58f. Added the comment at the call site: "zero means no limit; callers should treat a zero StalenessLimit as unlimited."

PR #14, opened 2026-09-16
feat(app): walking skeleton — single-downstream poll, transform, and serve

### LINKED ISSUE #3: Single-downstream poll, transform, and serve (walking skeleton)
## What to build

`New(Config) (*App, error)`, `(*App).Refresh(ctx) error`, and `(*App).Handler() http.Handler` working end to end against a single fake downstream (`httptest.Server`). `Refresh` fetches `GET {apiScheme}://{apiHost}:{apiPort}/api/rawdata`, filters routers by the downstream's entrypoint allow-list, excludes `internal`-provider and disabled routers, emits a router named `{downstream}-{router}` (provider qualifier stripped, sanitised to a safe character set) with the rule copied verbatim, entrypoints set to the configured edge entrypoints, and one service per downstream built from the configured traffic scheme/host/port with `passHostHeader: true`. `Handler` serves the merged result as JSON (`traefik/genconf` types, top-level `http` key, `routers`+`services` only) on the configured path with deterministic sorted-key serialisation and `ETag`/`If-None-Match` support, reading an atomically-swapped snapshot so an in-flight poll never blocks a request.

This is the tracer bullet: it proves the full path (fetch → transform → serve) for one downstream before multi-downstream merge, auth, failure handling, or observability are added.

## Acceptance criteria

- [ ] A `Refresh` against one fake downstream produces the expected router/service JSON when `Handler` is queried
- [ ] Internal-provider and disabled routers are excluded from the served output
- [ ] Routers without an allow-listed entrypoint are excluded
- [ ] Emitted router entrypoints are the edge's, not the downstream's
- [ ] Two `Refresh` cycles over unchanged fixtures produce byte-identical output and an unchanged `ETag`
- [ ] Tests use `httptest.Server` on both the fake downstream and the served handler; no mocked transport or injected clock

gringolito:
## Summary

Adds `internal/app` with three public symbols that carry a single downstream's Traefik routes to a file-provider JSON endpoint:

- `New(config.Config) (*App, error)`
- `(*App).Refresh(ctx) error`
- `(*App).Handler() http.Handler`

Also closes a latent gap in `internal/config`: `APIAddress` and `TrafficAddress` were validated only as non-empty. A malformed URL would have silently passed `config.Load` and blown up on the first `Refresh` call after deploy. `requireAbsoluteURL` now checks scheme and host at startup.

## Why

Issue #3 asks for a tracer bullet that proves the full path from fetch to serve for one downstream, before multi-downstream merge, auth, failure handling, or observability land.

## How it works

`Refresh` fetches `GET {APIAddress}/api/rawdata` and skips any router where:

- the provider is `internal` (checked against the `provider` field and the `@internal` key suffix)
- `status` is `disabled`
- none of its entrypoints appear in `Downstream.AllowedEntrypoints` (empty list means accept all)

For each router that passes, `Refresh` strips the `@provider` qualifier from the key, replaces characters outside `[a-zA-Z0-9_-]` with `-`, and prepends the downstream name. `my-router@docker` on downstream `primary` becomes `primary-my-router`. The output router carries `Config.EdgeEntrypoints`, not the downstream's original entrypoints, and points to a service named after the downstream. One service per downstream, `passHostHeader: true`, URL from `TrafficAddress`.

The result is marshalled using `github.com/traefik/genconf/dynamic` types (`dynamic.Configuration`, `dynamic.Router`, `dynamic.Service`, `dynamic.ServersLoadBalancer`) as the spec requires. Rawdata input types (`rawdataResponse`, `rawRouter`) stay local since they represent Traefik's internal API response format, which is not in genconf. `encoding/json` sorts map keys, so the body is deterministic. ETag is the first 8 bytes of SHA-256 over that body. Unchanged upstream data produces the same bytes and the same ETag across successive `Refresh` calls.

`Handler` stores the snapshot behind `atomic.Pointer` so reads never block an in-flight poll. It returns 304 on `If-None-Match` match and 503 before the first `Refresh`.

## Reviewer entry points

Start with `types.go` (17 lines, rawdata input shapes only), then `app.go` (`fetchAndMerge` for filtering and transform logic, `serveSnapshot` for ETag handling). The test file follows the same cycle numbering as the issue's acceptance criteria.

| File | Lines | What it contains |
|---|---|---|
| `internal/app/types.go` | 17 | Rawdata input types only; output uses genconf directly |
| `internal/app/app.go` | 191 | `New`, `Refresh`, `Handler`; `fetchAndMerge`, `serveSnapshot` |
| `internal/app/app_test.go` | 380 | Seven TDD cycles, one per AC; both sides use `httptest.Server` |
| `internal/config/config.go` | +`requireAbsoluteURL` | Absolute-URL validation added to `validate()` |
| `internal/config/config_test.go` | +`TestLoad_InvalidAPIURL` | Covers the new validation path |

`golangci-lint run ./...` reports 0 issues. `go test ./...` passes.

Closes #3


internal/app/app.go
```diff
@@ -0,0 +1,188 @@
...
+// configuration as a single Traefik file-provider JSON document.
+type App struct {
+	cfg    config.Config
+	client *http.Client
+	snap   atomic.Pointer[snapshot]
+}
+
+// New validates cfg and returns a ready App.  The App makes no network calls
+// until Refresh is invoked.
+func New(cfg config.Config) (*App, error) {
```
gringolito, line 39: `New`'s comment says it validates cfg, but no validation happens here — it lives in `config.Load`/`validate`. The `error` half of the return is also unreachable (the only path returns nil), yet every caller must handle it; the tests sidestep it with `a, _ :=`. Either validate here, drop the error return, or fix the comment. Judgement call: misleading name/comment, not a hard violation.

internal/app/app_test.go
```diff
@@ -0,0 +1,380 @@
...
+			fService:     vSvcDocker,
+			fRule:        vExampleRule,
+			fStatus:      vEnabled,
+			fProvider:    vDocker,
+		},
+	})
+	defer ds.Close()
+
+	cfg := testConfig(ds.URL, "http://traffic.example.com:80")
+	a, _ := app.New(cfg)
```
gringolito: This four-line shape (`testConfig` → `app.New` → `Refresh` → `queryHandler`) recurs in Cycles 3-7, and `a, _ :=` / `_ = a.Refresh` discard the errors Cycle 2 checks. A Refresh failure surfaces as a confusing assertion failure several lines later. Extract a `refreshAndQuery(t, ds)` helper that fatals on error. Judgement call: Duplicated Code.

internal/app/app.go
```diff
@@ -0,0 +1,188 @@
...
+	allow := entrypointSet(ds.AllowedEntrypoints)
+
+	for name, r := range raw.Routers {
+		if isInternalProvider(name, r.Provider) {
+			continue
+		}
+		if r.Status == "disabled" {
+			continue
+		}
+		if allow != nil && !hasAllowedEntrypoint(r.EntryPoints, allow) {
```
gringolito, line 108: Spec AC (issue #3): "Routers without an allow-listed entrypoint are excluded." With an empty allow-list, `entrypointSet` returns nil and this nil guard skips filtering entirely — an empty allow-list publishes every router. The PR body documents the deviation (empty = accept all), which is the more operationally sensible rule, but it diverges from the issue text and no test exercises an empty allow-list. Worth resolving in the issue before merge: amend the AC or add the test.

gringolito (commented):
Two-axis review (Standards / Spec) of `origin/main...HEAD` (merge-base `50af257`, commits `2c1344b` + `634828a`) against issue #3.

## Standards

No documented repo standards exist (no CONTRIBUTING.md / CODING_STANDARDS.md; README is one line), so there are no hard violations to cite. golangci-lint / format / `go vet` are tooling-enforced and were skipped. Findings are judgement calls:

- app.go:36 — comment claims `New` validates cfg, but validation lives in `config.Load`; the error return is unreachable yet every caller pays for it.
- app_test.go:181 — the same four-line setup shape recurs in Cycles 3-7, and `a, _ :=` / `_ = a.Refresh` discard the errors Cycle 2 checks, surfacing a Refresh failure as a confusing downstream assertion.
- app.go:71 — string-concat URL; a trailing slash in APIAddress yields a double slash (`url.JoinPath` would be robust). Borderline.

## Spec

- app.go:103 — empty allow-list publishes every router, diverging from AC "Routers without an allow-listed entrypoint are excluded". PR body documents it (empty = accept all), but nothing tests the empty case.
- types.go:23 — hand-rolled genconf subset rather than the real types; drops fields a file provider tolerates (`priority`, `tls`, `middlewares`). Shape-conformance is a defensible reading; confirm intent.
- app.go:179 — ETag compares the raw `If-None-Match` string; no `W/` weak-tag, list, or `*` handling (RFC 7232). Known gap, not required by the issue.
- config.go:194 — `requireAbsoluteURL` was not requested by issue #3; it silently tightens accepted config (bare `host:port` api_address now fails startup) inside a feature PR.
- Refresh failure path keeps the last good snapshot serving (correct for a skeleton that defers failure handling) but is untested — it falls inside codecov's 10 missing lines.

Verified conformant: `New`/`Refresh`/`Handler` wiring, rawdata fetch, internal-provider and disabled exclusion, sanitised `{downstream}-{router}` naming, rule copied verbatim, edge entrypoints, one service per downstream with `passHostHeader: true`, atomic snapshot swap, deterministic sorted JSON with stable ETag across cycles, 503 before first Refresh, and all six AC cycles running on `httptest.Server` on both sides.

Worst per axis — Standards: the misleading `New` contract (comment + dead error path). Spec: the entrypoint-filter divergence with the acceptance criterion.

internal/app/app.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (a *App) serveSnapshot(w http.ResponseWriter, r *http.Request) {
+	snap := a.snap.Load()
+	if snap == nil {
+		http.Error(w, "not ready", http.StatusServiceUnavailable)
+		return
+	}
+
+	if r.Header.Get("If-None-Match") == snap.etag {
```
gringolito, line 185: `If-None-Match` is compared as a raw string against the ETag. Per RFC 7232 the header can also be `W/"..."` (weak validator), a comma-separated list, or `*` — none handled here. Fine for the skeleton's happy path (the issue doesn't require more); flagging the gap, not a defect.

internal/app/types.go
```diff
@@ -0,0 +1,49 @@
...
+	// Status is "enabled", "disabled", or "warning".
+	Status   string `json:"status"`
+	Provider string `json:"provider"`
+}
+
+// The types below mirror github.com/traefik/genconf dynamic.Configuration,
+// restricted to the http.routers + http.services subset that traefik-scout
+// emits.
+
+type configOutput struct {
```
gringolito: Issue #3 says "traefik/genconf types". These mirror the `http.routers` + `http.services` subset but are private to this module, not the genconf module, and they drop fields a file provider tolerates (`priority`, `tls`, `middlewares`). Shape-conformance is a defensible reading for the tracer bullet — confirming intent: subset OK, or pull in the real types?

internal/config/config.go
```diff
@@ -190,11 +191,11 @@ func validate(cfg *Config) error {
 		if !validName.MatchString(d.Name) {
 			return fmt.Errorf("downstream[%d]: name %q contains invalid characters (allowed: a-z, A-Z, 0-9, _, -)", i, d.Name)
 		}
-		if d.APIAddress == "" {
-			return fmt.Errorf("downstream[%d] (%q): api_address is required", i, d.Name)
+		if err := requireAbsoluteURL("api_address", d.APIAddress); err != nil {
```
gringolito, line 194: Issue #3 doesn't ask for config validation changes. `requireAbsoluteURL` tightens `api_address`/`traffic_address` from "non-empty" to "absolute URL": a previously accepted bare `host:port` now fails startup. Mild but real scope creep riding in a feature PR — if it stays, it deserves its own commit or a changelog note, since it changes accepted config format.

internal/app/app.go
gringolito: Fixed. Comment now reads: "New returns a ready App configured from cfg. Config is already validated by config.Load; New always returns a nil error but keeps the error return for future validation at construction time." The spec requires the `(*App, error)` signature, so the error return stays.

internal/app/app_test.go
gringolito: Extracted `refreshAndQuery(t, ds)` helper that fatals on both `New` and `Refresh` errors. Applied to Cycles 3-6. Cycle 7 keeps its own setup since it needs two Refresh calls and direct access to the App.

internal/app/app.go
gringolito: Added `TestRefreshAndHandler_EmptyAllowListPassesAll` to document and pin the decision: empty `AllowedEntrypoints` passes all routers through. Added an inline comment at the nil guard in `fetchAndMerge` to make the intent explicit. The behavior is intentional (empty = no filter rather than exclude all), and the PR body already described it; the test makes it permanent.

internal/app/app.go
gringolito: Acknowledged. The current handler compares the raw `If-None-Match` string. Weak validators, lists, and `*` are all unhandled. Fine for this skeleton where the only client is the Traefik edge and we generate the ETag ourselves. Can be tightened in the observability or failure-resilience slice if needed.

internal/app/types.go
gringolito: Adopted in the latest commit. `types.go` now only contains rawdata input types (`rawdataResponse`, `rawRouter`). `app.go` imports `github.com/traefik/genconf/dynamic` directly and uses `dynamic.Configuration`, `dynamic.Router`, `dynamic.Service`, and `dynamic.ServersLoadBalancer`. Rawdata input types stay local since they represent Traefik's internal API response format, not part of genconf.

internal/config/config.go
gringolito: Fair point on scope. The validation was added after a review advisory called out that malformed addresses fail at first `Refresh` rather than startup, which is a real operational gap (deploy succeeds, service is silently broken). It lives in its own commit so the diff is isolated. If you'd prefer it pulled into a separate PR, I can revert the config changes here and open one against main.

PR #15, opened 2026-09-17
feat(app): add App with multi-downstream merge and collision detection

### LINKED ISSUE #4: Multi-downstream merge with collision detection and priority offsets
## What to build

Merge across two or more downstreams within a single `Refresh` cycle: every configured downstream is represented in the merged output regardless of response order or speed, router-name prefixing prevents two downstreams that both define a router with the same name from overwriting each other, per-downstream priority offsets are applied to emitted routers, and identical routing rules contributed by different downstreams are detected after merging and logged at warning level naming both contributors (the configuration is still served — Traefik itself resolves the conflict by priority).

## Acceptance criteria

- [ ] Three fake downstreams, each with distinct routes, all appear in one merged `Handler` response
- [ ] Two downstreams defining a router with the same name both appear under their prefixed names, not overwriting each other
- [ ] A configured priority offset shows up on the emitted router
- [ ] Two downstreams contributing an identical rule produce a warning-level log naming both downstream names, and the configuration is still served with both routers present

gringolito:
## Summary

Adds the `App` type (`New`/`Refresh`/`Handler`) that polls multiple downstream Traefik instances and serves their routes as a single merged dynamic configuration.

`Refresh` fetches each downstream's `/api/rawdata`, filters routers by entrypoint allow-list, drops disabled and internal-provider routers, prefixes every router name with the downstream name to prevent collisions, applies per-downstream priority offsets, and generates one service per downstream pointing at its configured traffic address. The merged JSON snapshot is atomically swapped so `Handler` never blocks on an in-flight poll.

After merging, if two downstreams contribute the same routing rule, `Refresh` logs a warning at `WARN` level naming both downstream contributors. The config is still served with both routers; Traefik resolves the conflict by priority.

Adds `github.com/traefik/genconf v0.5.2` for the output types so the JSON schema tracks Traefik's own definition.

## Why

Implements the core aggregation behavior described in issue #4: multi-downstream merge, router-name prefixing, priority offsets, and rule-collision detection.

All four acceptance criteria are covered by integration tests in `internal/app/app_test.go`. Tests drive real `httptest.Server` fake downstreams and assert on the JSON served by `app.Handler()`, with no mocked transport or internal seams.

Closes #4


internal/app/app.go
```diff
@@ -45,18 +46,34 @@ func New(cfg config.Config) (*App, error) {
...
+// the configuration is still served so Traefik can resolve the conflict by
+// priority.
 func (a *App) Refresh(ctx context.Context) error {
 	out := &dynamic.HTTPConfiguration{
 		Routers:  make(map[string]*dynamic.Router),
 		Services: make(map[string]*dynamic.Service),
 	}
 
+	// rule -> downstream names that contributed it (for collision detection)
+	ruleContributors := make(map[string][]string)
```
gringolito: Judgement call (Data Clumps / Divergent Change): this accumulator + `out` travel together as parallel out-params through `fetchAndMerge`, and the provenance added at L138 is largely derivable after the loop from `out.Routers` (prefixed keys → routers carrying `.Rule`), so collisions could be computed in a post-pass without the extra parameter. Not wrong as written; consider bundling `out` + provenance into a merge-result type if collision detail grows.

internal/app/app.go
```diff
@@ -45,18 +46,34 @@ func New(cfg config.Config) (*App, error) {
...
+
 	for _, ds := range a.cfg.Downstreams {
-		if err := a.fetchAndMerge(ctx, ds, out); err != nil {
+		if err := a.fetchAndMerge(ctx, ds, out, ruleContributors); err != nil {
```
gringolito: Spec partial against issue #4: "every configured downstream is represented in the merged output regardless of response order or speed." This loop is sequential and fail-fast: any downstream fetch error returns and aborts `Refresh`, so no snapshot is replaced and no downstream is served. Order-independence holds (prefixing), but one slow/failing downstream blocks or kills the merge for all. The error path is untested. Consider pinning intended semantics in an issue: concurrent fetch, or skip-on-error with the healthy downstreams still served.

internal/app/app.go
```diff
@@ -45,18 +46,34 @@ func New(cfg config.Config) (*App, error) {
...
 			return err
 		}
 	}
 
+	for rule, contributors := range ruleContributors {
+		if len(contributors) > 1 {
+			slog.Warn("identical routing rule detected",
```
gringolito: Package-level `slog.Warn` instead of an injected `*slog.Logger`. The test observing this warning has to swap global state (`slog.SetDefault` + manual restore), which is brittle under parallel tests. Injecting a logger would let tests pass one and assert on its output.

internal/app/app.go
```diff
@@ -70,7 +87,7 @@ func (a *App) Refresh(ctx context.Context) error {
 	return nil
 }
 
-func (a *App) fetchAndMerge(ctx context.Context, ds config.Downstream, out *dynamic.HTTPConfiguration) error {
+func (a *App) fetchAndMerge(ctx context.Context, ds config.Downstream, out *dynamic.HTTPConfiguration, ruleContribs map[string][]string) error {
```
gringolito: Mysterious Name (mild): this parameter is `ruleContribs` while the declaration in `Refresh` is `ruleContributors`. Same concept, two names in one call chain; pick one.

internal/app/app.go
```diff
@@ -116,7 +133,9 @@ func (a *App) fetchAndMerge(ctx context.Context, ds config.Downstream, out *dyna
 			EntryPoints: a.cfg.EdgeEntrypoints,
 			Service:     ds.Name,
 			Rule:        r.Rule,
+			Priority:    r.Priority + ds.PriorityOffset,
 		}
+		ruleContribs[r.Rule] = append(ruleContribs[r.Rule], ds.Name)
```
gringolito: Nit: collision detection keys on the raw `r.Rule` string, so `Host("a.example.com")` vs `Host("a.example.com" )` would not be detected as the same rule. Harmless for the current fixtures; a normalized key would be more robust.

internal/app/app_test.go
```diff
@@ -68,6 +72,28 @@ func fakeDownstream(routers map[string]any) *httptest.Server {
 	}))
 }
 
+// fakeDownstreamJSON starts an httptest.Server serving a raw JSON string at
+// /api/rawdata.  Use when the fixture body is easier to express as a string.
+func fakeDownstreamJSON(body string) *httptest.Server {
```
gringolito: Duplicated Code (mild): this handler re-implements the skeleton of the pre-existing `fakeDownstream` (same path check, 404 branch, Content-Type header, write), differing only in how the body is produced. Consider a shared handler shape.

internal/app/app_test.go
```diff
@@ -410,3 +436,173 @@ func TestRefreshAndHandler_EmptyAllowListPassesAll(t *testing.T) {
...
+// merged Handler response.
+func TestRefresh_ThreeDownstreams_AllRoutesAppear(t *testing.T) {
+	ds1 := fakeDownstreamJSON(rawDataJSON("app-a", "web", `Host("a.example.com")`, "docker"))
+	ds2 := fakeDownstreamJSON(rawDataJSON("app-b", "web", `Host("b.example.com")`, "docker"))
+	ds3 := fakeDownstreamJSON(rawDataJSON("app-c", "web", `Host("c.example.com")`, "docker"))
+	defer ds1.Close()
+	defer ds2.Close()
+	defer ds3.Close()
+
+	cfg := config.Config{
```
gringolito: Duplicated Code: this `config.Config{...}` literal (ConfigPath, RequestTimeout, MaxResponseSize, EdgeEntrypoints, Downstreams) is repeated verbatim in all four new tests. Extract a helper that takes only the varying `Downstreams` slice (and entrypoints).

gringolito (commented):
Two-axis review of `5ab0f8f...320d614` (fixed point: PR base `origin/main`) against issue #4. No documented coding standards exist in this repo (no CODING_STANDARDS.md/CONTRIBUTING.md; README is a one-liner), so there are zero hard violations. The Standards findings are judgement calls against a code-smell baseline; tooling-enforced lint is excluded.

## Standards (5 findings, all judgement calls)

1. **internal/app/app.go** — the collision accumulator is `ruleContributors` in `Refresh` but renamed `ruleContribs` as the `fetchAndMerge` parameter. One concept, two names in a single call chain. (Mysterious Name)
2. **internal/app/app.go** — `fetchAndMerge` gains a second out-param: `out` and `ruleContribs` travel together (Data Clumps), and the provenance recorded at L138 is largely derivable from `out.Routers` after the loop (prefixed keys map to routers carrying `.Rule`), so `Refresh` could compute collisions in a post-pass without the extra signature. If collision detail grows, bundle `out` + provenance into a merge-result type. (Divergent Change)
3. **internal/app/app.go L70** — package-level `slog.Warn` instead of an injected `*slog.Logger`. The new test must mutate global state (`slog.SetDefault` with manual restore) to observe the warning, which is brittle under parallel tests. The cost lands in this same diff's test code.
4. **internal/app/app_test.go** — the ~9-line `config.Config{...}` literal is repeated verbatim in all four new tests. Extract a helper taking only the varying `Downstreams` slice. (Duplicated Code)
5. **internal/app/app_test.go** — `fakeDownstreamJSON` re-implements the handler skeleton of the pre-existing `fakeDownstream` (same path check, 404 branch, Content-Type header, write), differing only in body production. Share one handler shape. (Duplicated Code)

## Spec (issue #4): PASS with one partial and two nits

All four acceptance criteria are implemented and covered by behavioral tests (real `httptest.Server` fakes, assertions on the JSON served by `Handler`):

- Three downstreams merged into one `Handler` response.
- Router names always prefixed with the downstream name, so two same-named routers cannot overwrite each other.
- Per-downstream `PriorityOffset` applied (`r.Priority + ds.PriorityOffset`).
- Identical rules detected post-merge and logged at `WARN` naming both contributors, with both routers still served.

- **Partial** — "every configured downstream is represented in the merged output regardless of response order or speed": `Refresh` is sequential and fail-fast (see inline comment on the fetch loop). Order-independence holds, but one slow or failing downstream aborts the merge for all downstreams. Error path is untested. Recommend an issue to pin intended semantics (concurrent fetch / skip-on-error).
- **Nit** — collision detection keys on the raw rule string, so it is whitespace-sensitive.
- **Nit** — emitting `priority: 0` for zero offsets is harmless.

No scope creep. `go.mod`/`go.sum`/`types.go` changes are clean.

internal/app/app.go
gringolito: Acknowledged. The parallel out-params are a mild smell here; noted. If collision detail grows (e.g. per-router context or severity levels), bundling `out` and provenance into a named result type would be the clean next step. Deferred to a follow-up rather than pre-optimising now.

internal/app/app.go
gringolito: Opened #16 to pin the intended semantics (concurrent fetch vs. skip-on-error vs. current fail-fast). The spec says individual failures should not block healthy downstreams, so this is a real gap — just not in scope for this AC. Will address in #16.

internal/app/app.go
gringolito: Fixed. Added `Option`/`WithLogger(*slog.Logger)` functional option to `New`. `App` now stores the logger and calls `a.log.Warn`. The collision test passes `app.WithLogger(logger)` directly — no global state mutation.

internal/app/app.go
gringolito: Fixed. Parameter renamed to `ruleContributors` throughout `fetchAndMerge`, matching the variable name in `Refresh`.

internal/app/app.go
gringolito: Agreed, it is whitespace-sensitive. A normalised key (trimming and collapsing internal whitespace) would be more robust. Left as-is for now since real Traefik rule strings are machine-generated and don't carry incidental whitespace in practice. Happy to add normalisation if you want it before merge.

internal/app/app_test.go
gringolito: Fixed. Removed `fakeDownstreamJSON` and `rawDataJSON`. All four new tests now use `fakeDownstream` with the same `map[string]any` fixture style as the existing suite.

internal/app/app_test.go
gringolito: Fixed. Extracted `multiConfig(downstreams []config.Downstream) config.Config` helper used by all four new tests. The only varying input per test is the `Downstreams` slice.

internal/app/app.go
gringolito: Close #16 and address it now

internal/app/app.go
gringolito: Not a judgment call anymore, I want you to do this.

internal/app/app.go
gringolito: Done in 54cccfb. `Refresh` now logs at WARN and continues on a downstream error instead of aborting. Healthy downstreams are always merged and served. Added `TestRefresh_FailingDownstream_HealthyDownstreamsStillServed` to cover the path. Closed #16.

internal/app/app.go
gringolito: Done in 54cccfb. Introduced `mergeResult` (unexported struct bundling `out *dynamic.HTTPConfiguration` and `ruleContributors map[string][]string`). `fetchAndMerge` now takes `*mergeResult` as a single accumulator instead of two parallel out-params.

Issue #2, opened 2026-09-16
Spec: Traefik multi-instance config aggregator (traefik-scout)


gringolito:
Source spec: gringolito/portainer-stacks#1

## Problem Statement

I run several independent Traefik instances across my homelab, and I want one public entry point in front of all of them.

`*.${TRAEFIK_PUBLIC_DOMAIN}` resolves to a single IP, so every request has to arrive at one edge Traefik. That edge instance owns the wildcard certificate and terminates TLS once. Behind it sit multiple Traefik instances, each watching its own Docker daemon: one on my general-purpose Docker host, one on the GPU host, one serving ingress inside Kubernetes, and eventually an entire non-prod network.

The edge cannot discover those routes on its own. Traefik permits only one `docker` provider in a static configuration, so it can't watch more than one remote daemon even if I exposed their sockets.

Existing community plugins that claim to solve this do not work for my case.

I also refuse to let my lab Docker host reach into my other networks. Any solution has to be initiated from the edge; the edge dials into each network, and no network dials out.

## Solution

A small, standalone Go service that runs beside the edge Traefik, polls each downstream Traefik's API, rewrites what it finds into a single merged dynamic configuration, and serves that over HTTP for the edge to consume through its built-in `http` provider.

Each downstream keeps doing what it already does: watch its local Docker daemon, build routes from container labels, and apply its own middlewares. The aggregator copies each downstream router's matching rule and repoints its service at that downstream's web entrypoint. Traffic then flows edge → downstream Traefik → container, so downstream middlewares execute naturally where they were defined, and I never have to replicate them at the edge.

Because the aggregator initiates every connection outward from the edge, my network policy stays one-directional. Because it speaks only to Traefik's API and never to a Docker socket, the blast radius of a compromise is a list of hostnames rather than root on a container host. And because it's an ordinary Go binary rather than a Yaegi-interpreted plugin, it can be tested, debugged, and instrumented like normal software.

Deploying a new service anywhere in the homelab becomes: add labels to the container, as I already do. The route appears at the edge within one poll interval.

## User Stories

1. As a homelab operator, I want one public entry point in front of several independent Traefik instances, so that a single wildcard DNS record and a single certificate cover every service I host.
2. As a homelab operator, I want new services to appear at the edge automatically after I label a container, so that I never hand-maintain a route list.
3. As a homelab operator, I want to add a new downstream Traefik by adding one entry to a configuration file, so that onboarding a host or a whole network is a small, obvious change.
[...]

PR #17, opened 2026-09-17
feat(app): per-downstream auth and TLS for poll requests

### LINKED ISSUE #5: Per-downstream authentication and TLS
## What to build

Per-downstream authentication on the poll request: optional static headers (bearer tokens, API keys), optional HTTP basic auth, and a TLS block for the API connection (custom CA, client certificate + key, insecure-skip-verify). Asserted by a fake downstream server that inspects the credentials/TLS state it actually received, not by inspecting aggregator internals.

## Acceptance criteria

- [ ] A downstream configured with a bearer/API-key header sends that exact header on every poll request
- [ ] A downstream configured with basic auth sends valid HTTP basic auth credentials
- [ ] A downstream configured with a custom CA successfully polls a server presenting a certificate signed by that CA
- [ ] A downstream configured with insecure-skip-verify successfully polls a server with a self-signed certificate
- [ ] A downstream configured with a client certificate presents it during the TLS handshake

gringolito:
## What changed

`New` builds one `http.Client` per downstream instead of one shared client. Each client carries a transport configured with the downstream's TLS settings: custom CA pool, client certificate, and insecure-skip-verify. Auth is applied per request in `fetchRawData`: arbitrary static headers covering bearer tokens and API keys, plus optional HTTP basic auth.

`Auth.token` is gone. Use `auth.headers` instead, a free-form map:

```yaml
auth:
  headers:
    Authorization: "Bearer mytoken"     # bearer
    X-Api-Key: "mykey"                  # API key
```

`auth.headers` and `auth.username`/`password` are mutually exclusive. Two other startup validation rules: `username` and `password` must both be set or both be absent; `cert` and `key` must be set together. All three fail at load time with a message naming the fields.

## Where to start

`internal/app/app.go` has two focused changes. `buildClient` (lines 79-116) builds the per-downstream client, cloning `http.DefaultTransport` and overriding only `TLSClientConfig` to preserve proxy and HTTP/2 settings. `fetchRawData` picks `a.clients[ds.Name]` and applies headers and basic auth before the request.

Config types were already present in `internal/config/config.go`. `Auth.Token` is replaced by `Auth.Headers map[string]string`. `validate()` now checks the three pairing rules.

`internal/app/app_test.go` has six `TestRefresh_*` tests for Issue #5: one per acceptance criterion, plus an explicit API-key arbitrary-header test. Two shared helpers, `rawdataHandler` and `serveRawdata`, handle the path routing and JSON serialisation that all fake downstream servers need. The TLS tests generate ephemeral ECDSA P-256 certs at runtime, no fixtures on disk.

`internal/config/config_test.go` has six new tests: headers YAML parsing, both auth pairing errors, both TLS pairing errors, and the headers/basic-auth mutual exclusion.

Closes #5


internal/app/app.go
```diff
@@ -151,7 +200,16 @@ func (a *App) fetchRawData(ctx context.Context, ds config.Downstream) (*rawdataR
 		return nil, fmt.Errorf("build request for %s: %w", ds.Name, err)
 	}
 
-	resp, err := a.client.Do(req)
+	if ds.Auth != nil {
+		switch {
+		case ds.Auth.Token != "":
```
gringolito: **[Standards/Spec]** `config.Auth` permits Token and Username/Password together (config.go:49-53); when both are set, the token silently wins and basic auth is dropped. Deterministic but undisclosed. Reject the combination in `config.Load`, or document the precedence on the `Auth` type.

internal/app/app.go
```diff
@@ -151,7 +200,16 @@ func (a *App) fetchRawData(ctx context.Context, ds config.Downstream) (*rawdataR
 		return nil, fmt.Errorf("build request for %s: %w", ds.Name, err)
 	}
 
-	resp, err := a.client.Do(req)
+	if ds.Auth != nil {
```
gringolito, line 205: **[Spec]** Issue #5 asks for "optional static headers (bearer tokens, API keys)", but `config.Auth` (config.go:49-53) only carries `Token` / `Username` / `Password`. AC 1 "sends that exact header" is satisfiable only for `Authorization: Bearer`; a downstream expecting e.g. `X-API-Key` cannot be configured. Either extend `Auth` with a headers map, or narrow the issue wording.

internal/app/app.go
```diff
@@ -64,6 +76,43 @@ func New(cfg config.Config, opts ...Option) (*App, error) {
...
+		}
+		pool := x509.NewCertPool()
+		if !pool.AppendCertsFromPEM(caPEM) {
+			return nil, fmt.Errorf("CA file %q: no valid PEM certificates found", ds.TLS.CA)
+		}
+		tlsCfg.RootCAs = pool
+	}
+
+	if ds.TLS.Cert != "" || ds.TLS.Key != "" {
+		cert, err := tls.LoadX509KeyPair(ds.TLS.Cert, ds.TLS.Key)
```
gringolito, line 103: **[Standards]** If `key` is set without `cert` (or vice versa), this fails at startup with a low-level `tls.LoadX509KeyPair` error instead of a validation message naming the field. Minor: consider a config-time "cert and key must both be set" check in `config.Load`.

internal/app/app_test.go
```diff
@@ -739,3 +757,277 @@ func TestRefresh_AllDownstreamsFail_LastGoodRoutesRetained(t *testing.T) {
...
+	return f.Name()
+}
+
+// requestCheckingDownstream starts a fake downstream that calls checkFn on
+// every /api/rawdata request before serving an empty rawdata JSON response.
+func requestCheckingDownstream(t *testing.T, checkFn func(t *testing.T, r *http.Request)) *httptest.Server {
+	t.Helper()
+	body, _ := json.Marshal(map[string]any{jsonFieldRouters: map[string]any{}})
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != rawdataPath {
```
gringolito: **[Standards]** Duplicated Code: this "404 unless /api/rawdata, then serve JSON" handler body is copy-pasted five times — `fakeDownstream` (:110-117), `sequencedDownstream` (:125-131), here (:881-889), `tlsDownstreamWith` (:898-905), and inline in `TestRefresh_InsecureSkipVerify` (:992-1000). Extract one helper (e.g. route all five through `requestCheckingDownstream`) so a fake change is made once.

internal/app/app.go
```diff
@@ -64,6 +76,43 @@ func New(cfg config.Config, opts ...Option) (*App, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("load client certificate: %w", err)
+		}
+		tlsCfg.Certificates = []tls.Certificate{cert}
+	}
+
+	return &http.Client{
+		Timeout:   timeout,
+		Transport: &http.Transport{TLSClientConfig: tlsCfg},
```
gringolito: **[Spec]** TLS downstreams get a bare `&http.Transport{TLSClientConfig: tlsCfg}` here, which drops proxy-from-environment and HTTP/2 behaviour that the no-TLS path keeps via `http.DefaultTransport` (line 83). Behaviour now diverges between TLS and non-TLS downstreams beyond TLS itself. If unintended, clone `http.DefaultTransport` and set `TLSClientConfig` on it.

gringolito (commented):
Two-axis review of `d9f9502...f717923` (merge-base = main) against issue #5. Inline comments are linked where they land.

## Standards

**Hard violations: none.** The diff respects every rule `.golangci.yaml` enforces (gosec sites carry `//nolint:gosec` with reasons, `testpackage` external test package, `%w` error wrapping, `goconst` shared consts). The repo documents no prose standards file, so the smell baseline (Fowler, _Refactoring_ ch.3) applies.

Judgement calls:

1. **Duplicated Code** — the "404 unless `/api/rawdata`, then serve JSON" handler is copy-pasted five times: `fakeDownstream` (app_test.go:110-117), `sequencedDownstream` (:125-131), `requestCheckingDownstream` (:881-889), `tlsDownstreamWith` (:898-905), and inline in `TestRefresh_InsecureSkipVerify` (:992-1000). Extract one helper; a future fake change (content-type, query checks) currently needs five edits. [inline](https://github.com/gringolito/traefik-scout/pull/17#discussion_r4032599890)
2. **Undisclosed credential precedence** — `config.Auth` permits token + basic together; when both are set the token silently wins and basic auth is dropped. Deterministic but undocumented; reject the combination in `config.Load` or document precedence. [inline](https://github.com/gringolito/traefik-scout/pull/17#discussion_r4032599844)
3. **Half-specified client cert** — `key` without `cert` (or vice versa) fails at startup with a low-level loader message rather than a validation error naming the field. [inline](https://github.com/gringolito/traefik-scout/pull/17#discussion_r4032599878)

Test CA/cert helpers (`newTestCA`, `signServerCert`, `signClientCert`) are well-factored; no duplication there.

## Spec

Spec source: issue #5. All 5 acceptance criteria are implemented and asserted via fake downstreams that inspect what the server actually received — headers via `requestCheckingDownstream`, TLS state via the handshake itself (`tlsDownstreamWith` + `RequireAndVerifyClientCert`), matching the issue's "inspects the credentials/TLS state it actually received" requirement: bearer (app_test.go:904-919), basic auth (:924-942), custom CA (:946-961), insecure-skip-verify (:965-1001), client cert (:1005-1032).

Findings:

1. **Missing: static-header / API-key support** — the issue asks for "optional static headers (bearer tokens, API keys)", but `config.Auth` has only `Token`/`Username`/`Password`, so AC 1 is satisfiable only for `Authorization: Bearer`. A downstream expecting `X-API-Key` cannot be configured. [inline](https://github.com/gringolito/traefik-scout/pull/17#discussion_r4032599859)
2. **Partial: "every poll request"** — headers are set per-request in `fetchRawData` (structurally satisfied), but each test exercises a single poll per Refresh, so "every poll" is never asserted across repeated cycles. Minor.
3. **Untested fail-fast path** — the PR body claims fail-fast on unreadable TLS files; `New` wraps `buildClient` errors (app.go:58-78), but no test covers an unreadable CA/key file or malformed PEM. This is the 4-line codecov gap.
4. **Scope creep (minor)** — pre-existing test literals refactored to shared consts (`rawdataPath`, `jsonFieldRouters`, `pemTypeECKey`); harmless but outside issue #5 and inflates the diff.

Suspect:

5. **TLS transport diverges from default** — TLS downstreams get a bare `&http.Transport{TLSClientConfig: tlsCfg}` (app.go:112), dropping proxy-from-environment and HTTP/2 that the no-TLS path keeps via `http.DefaultTransport`. Spec-compliant, but confirm the divergence is intended. [inline](https://github.com/gringolito/traefik-scout/pull/17#discussion_r4032599903)
6. **Token/basic precedence unpinned** — same hunk as Standards #2; the spec asks for "optional" auth, no precedence rule, and no test pins it.

Summary: **Standards** 3 findings, all judgement calls — worst is the 5× duplicated test handler. **Spec** 4 findings + 2 suspects — worst is API-key/static-header support absent from the issue text.

internal/app/app.go
gringolito: I prefer rejecting it

internal/app/app.go
gringolito: Extend Auth

internal/app/app.go
gringolito: Why this wasn't addressed?

internal/app/app.go
gringolito: Fixed. `auth.headers` and `auth.username`/`password` are now mutually exclusive: `validate()` returns an error when both are set, naming both fields. The original silent-win behaviour is gone entirely since `Auth.Token` was replaced by `Auth.Headers` in the same pass.

internal/app/app.go
gringolito: `Auth.Token` is replaced by `Auth.Headers map[string]string`. Any header name works now — `Authorization: Bearer X` for bearer tokens, `X-Api-Key: key` for API keys, or any other static header. Two tests cover it: `TestRefresh_BearerToken_SendsAuthorizationHeader` and `TestRefresh_APIKeyHeader_SendsCustomHeader`.

internal/app/app.go
gringolito: Fixed in an earlier commit. `validate()` now checks both directions: `cert` without `key` and `key` without `cert` both produce a startup error naming the fields. Four config tests cover each direction of each rule.

internal/app/app_test.go
gringolito: Fixed. Two shared helpers now handle the duplication: `rawdataHandler` wraps a handler func behind the path routing guard, and `serveRawdata` encodes the JSON body. All five sites — `fakeDownstream`, `sequencedDownstream`, `requestCheckingDownstream`, `tlsDownstreamWith`, and the inline handler in the insecure-skip-verify test — now delegate to these. The insecure-skip-verify test was simplified to call `tlsDownstreamWith(t, &tls.Config{})` directly.

internal/app/app.go
gringolito: Fixed. `buildClient` now calls `http.DefaultTransport.(*http.Transport).Clone()` and sets `TLSClientConfig` on the clone. Proxy-from-environment, connection pool settings, and HTTP/2 are all preserved; only the TLS config differs from the default.

PR #19, opened 2026-09-17
feat(app): per-downstream failure isolation, backoff, and staleness

### LINKED ISSUE #6: Downstream failure resilience and staleness
## What to build

Per-downstream failure isolation and staleness handling inside `Refresh`. Each downstream keeps its own last-known-good snapshot: a 500 response, an unreachable host, malformed JSON, or an oversized response from one downstream never blanks its own previously-served routes nor affects any other downstream's result in the same cycle. Failed polls are retried with capped exponential backoff. A downstream that returns HTTP 200 with zero routers is a valid empty result and replaces its snapshot with an empty one (not treated as a failure). An optional per-downstream staleness limit (disabled by default) withdraws that downstream's routes once no poll has succeeded for the configured duration; a downstream that recovers after failing has its routes return without a restart.

## Acceptance criteria

- [ ] One downstream returning 500 while others succeed leaves the healthy downstreams' routes intact and retains the failing downstream's last-known-good routes
- [ ] A downstream that becomes unreachable mid-run keeps serving its last successful snapshot
- [ ] A downstream returning zero routers replaces its snapshot with empty, without being logged/counted as a failure
- [ ] Malformed JSON and an over-size response from a downstream are handled without crashing `Refresh` or corrupting other downstreams' output
- [ ] With staleness limit enabled and exceeded, a persistently-dead downstream's routes are withdrawn from the merged output; with it disabled (default), routes are retained indefinitely
- [ ] A downstream that recovers after failing has its routes reappear on the next successful `Refresh`, no restart required

gringolito:
## Summary

Each downstream in `Refresh` now tracks its own retry state instead of sharing one pass/fail outcome for the whole poll cycle. A failing downstream backs off with a capped exponential window (1s doubling to 5m) instead of retrying every cycle, and no longer affects any other downstream's result. An optional per-downstream staleness limit withdraws a persistently-dead downstream's routes from the merged output instead of serving last-known-good data forever; recovery needs no restart.

## Why

A 500, unreachable host, malformed JSON, or oversized response from one downstream was already isolated from the others in `Refresh`, but that isolation had no test coverage (the Codecov gap flagged on #14) and the failure path retried on every cycle regardless of how long the downstream had been down. This adds the missing backoff and staleness behavior and closes the coverage gap with regression tests for each failure mode.

Closes #6


internal/app/resilience_test.go
```diff
@@ -0,0 +1,505 @@
...
+
+// Issue #6 tests that need to control the passage of time live in the internal
+// test package so they can inject a fake clock through the unexported
+// withClock Option without exposing a clock knob in the public API.
+
+// test-local copies of fixture constants defined in the external test package.
+const (
+	rawdataPath      = "/api/rawdata"
+	jsonFieldRouters = "routers"
+	testWeb          = "web"
```
gringolito: **Standards / possible Duplicated Code (judgement call).** `rawdataPath`, `jsonFieldRouters`, and `testWeb` re-declare fixture constants that already exist under different names in `app_test.go`. The file's own comment explains this is forced by needing package-internal access for `withClock`, which is a legitimate constraint — but consider whether the handful of shared literals (`"routers"`, `"web"`, `"/api/rawdata"`) could live in a small internal-only helper file shared by both `_test.go` files in package `app`, rather than living twice.

internal/app/app.go
```diff
@@ -142,21 +195,80 @@ func newMergeResult() *mergeResult {
...
 // served so Traefik can resolve the conflict by priority.
 func (a *App) Refresh(ctx context.Context) error {
 	result := newMergeResult()
 	contributed := false
+	// withdrew records that a staleness limit removed previously-served data
+	// this cycle: the snapshot must be republished (possibly empty) even when
+	// no downstream contributed, so withdrawal takes effect.
+	withdrew := false
 
 	for _, ds := range a.cfg.Downstreams {
```
gringolito, line 200: **Standards / possible Divergent Change (judgement call).** The `Refresh` loop body now interleaves four concerns in one block: backoff-window gating, network fetch + failure bookkeeping, staleness evaluation, and merge contribution. It's still readable at this size, but if a third policy axis is added later, consider extracting backoff/staleness decisions onto `pollState` methods (e.g. `st.dueBackoff(now)`, `st.isStale(now, limit)`) so `Refresh` just orchestrates rather than computing.

internal/app/app.go
```diff
@@ -142,21 +195,80 @@ func newMergeResult() *mergeResult {
...
+
+		// Enforce the optional per-downstream staleness limit: if no poll has
+		// succeeded within the limit, withdraw this downstream's routes for
+		// this cycle instead of serving last-known-good data indefinitely.
+		// This applies to fallbacks from failures AND to cycles deferred by
+		// backoff, and uses the last-success timestamp (never refreshed by
+		// either).
+		if raw != nil && ds.StalenessLimit > 0 {
+			if st.lastSuccess.IsZero() || a.now().Sub(st.lastSuccess) > ds.StalenessLimit {
+				a.log.WarnContext(ctx, "downstream stale, withdrawing routes",
```
gringolito: **Spec — unbounded log volume for a persistently-stale downstream.** This `WarnContext` fires on every `Refresh` cycle for as long as a downstream stays past its staleness limit, including cycles that are also skipped by backoff (per the comment above at line 261-262, the staleness check deliberately still runs during backoff skips). Unlike the failure-poll warning at line 246 — whose frequency is naturally throttled by the backoff window itself — this log call has no such throttle, so a permanently-dead downstream with a staleness limit configured will emit one WARN per poll interval (every `PollInterval`, default 30s) indefinitely. Consider only logging on the transition into withdrawal (e.g. when `raw` was non-nil going into this check, or by tracking a `st.withdrawn` flag) rather than on every cycle the downstream remains stale.

gringolito (commented):
Two-axis review (Standards + Spec) against issue #6.

## Standards
No documented CODING_STANDARDS.md/CONTRIBUTING.md exists in this repo, so the only standards source is the Fowler smell baseline (judgement calls only). Both findings below are judgement calls, not hard violations, and CI's golangci-lint is green. One-line summary: 2 low-priority judgement-call smells, worst is the growing responsibility list inside `Refresh` (still readable, not a blocker).

## Spec
All of issue #6's acceptance criteria are implemented and covered by tests: per-downstream failure isolation (500s, unreachable hosts, malformed JSON, oversized responses), capped exponential backoff, zero-routers-as-valid-empty-result, optional staleness withdrawal, and restart-free recovery. No missing/partial requirements and no unrequested scope creep (the `withClock` seam and backoff mechanism are both required to implement what the issue actually asked for). One-line summary: 1 finding, a log-volume issue in the staleness-withdrawal path (see inline comment on app.go) — otherwise Spec passes.

Overall: no blocking bugs found; findings below are non-blocking quality/operational suggestions.

internal/app/resilience_test.go
gringolito: I upvote this

gringolito: Fixed: added `internal/app/export_test.go` exposing `WithClock`/`WithSleep` test-only hooks, and moved `resilience_test.go` from `package app` to `package app_test`. It now reuses `app_test.go`'s existing `rawdataPath`, `jsonFieldRouters`, and `valueWeb` constants instead of re-declaring them; the `//nolint:testpackage` directive is gone. Verified zero duplicate fixture constants remain (commit b6d8602).

internal/app/app.go
gringolito: Addressed as part of the retry redesign below rather than a standalone extraction: `Refresh`'s per-downstream loop now delegates the fetch+retry decision to `fetchWithRetry`, and the staleness decision is gated on the `pollState.stale` transition, so the loop body reads as fetch-or-retry, then check staleness, then merge, instead of interleaving backoff bookkeeping inline. Left as three sequential steps rather than fully separate methods since the loop is still short and single-purpose after the backoff logic moved out (commit b6d8602).

internal/app/app.go
gringolito: Fixed: `pollState` now tracks a `stale` bool. The staleness WARN fires only on the false→true transition into withdrawal (plus a one-time INFO log on recovery), not once per cycle. `TestRefresh_Staleness_WarnsOnlyOnTransition` proves exactly one WARN across 4 consecutive stale cycles, zero on recovery, and exactly one more if it goes stale again (commit b6d8602).

gringolito:
Follow-up fix (not from the posted review, caught during self-review): the original backoff design skipped a failing downstream's next poll via a cross-cycle `notBefore` timestamp, but this repo has no poll loop yet (`Refresh` is only called from tests), so that gating didn't actually engage until several minutes of consecutive failures had accumulated, and the doc comment describing it was inaccurate. Replaced it with bounded in-call retries (up to 3 attempts, capped exponential delay ~500ms-2s) entirely inside one `Refresh` invocation, so the retry behavior in the issue's spec ("failed polls are retried with capped exponential backoff") is self-contained and doesn't depend on a scheduler this repo doesn't have yet. See commit b6d8602.

internal/app/app.go
```diff
@@ -30,6 +30,32 @@ type snapshot struct {
...
 }
 
+// Fetch-retry timing for a failing downstream, applied entirely within a
+// single Refresh cycle so resilience does not depend on any external poll
+// cadence.  A failing downstream is attempted fetchRetryMaxAttempts times,
+// sleeping fetchRetryBaseDelay (doubled per attempt, capped at
+// fetchRetryMaxDelay) between attempts.  500ms doubling to 2s rides out
+// brief blips (connection resets, one bad response) while bounding the
+// worst-case added latency for a permanently-dead downstream at about 1.5s
+// per cycle, small enough to sit inside a normal poll budget.
```
gringolito:
```suggestion
// fetchRetryMaxDelay) between attempts.
```

Noisy, cut this off

internal/app/app.go
```diff
@@ -39,6 +59,19 @@ func WithLogger(l *slog.Logger) Option {
...
 
+// withClock overrides the App's time source.  It is unexported on purpose:
+// only tests inject a clock; production always uses time.Now.
+func withClock(now func() time.Time) Option {
+	return func(a *App) { a.now = now }
+}
+
+// withSleep overrides the delay mechanism between fetch-retry attempts.  It
+// is unexported on purpose: only tests inject a sleep; production always
+// uses time.Sleep.
```
gringolito: Okay, this is something that concerns me. The tests should be black-boxed, and we shouldn't have been using un-exported or internal features to test the package. If we need time to be injected, let's refactor it to use dependency injection, not patch things on the fly.

internal/app/export_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+// Test-only exports of unexported App hooks, so the external test package
+// (app_test) can reach them without widening the production API.  Every
+// identifier here is referenced only from _test.go files.
+
+// WithClock exposes withClock to the app_test package.
+var WithClock = withClock
+
+// WithSleep exposes withSleep to the app_test package.
+var WithSleep = withSleep
```
gringolito: Hacky tests. Don't do it. Refactor to use dependency injection instead.

internal/app/app.go
gringolito: Agreed, fixed: removed the export_test.go shim entirely. WithClock and WithSleep are now ordinary exported Options on the public API, right next to WithLogger, using the same functional-options pattern the package already had. Tests construct the App through app.New(cfg, app.WithClock(...), app.WithSleep(...)) like any other caller would; no internal-package access or test-only aliasing left (commit 3184377).

internal/app/export_test.go
gringolito: Fixed: deleted export_test.go and exported WithClock/WithSleep as real Options (commit 3184377). Tests now inject the clock and sleep hook through the public constructor the same way WithLogger already worked, no more internal-package test file.

PR #20, opened 2026-09-17
feat(app): add health, readiness, metrics, and structured logs

### LINKED ISSUE #7: Observability: health, readiness, metrics, structured logs
## What to build

Observability surface served through `Handler`: `/healthz` returns 200 whenever the process is serving, independent of downstream health. `/readyz` returns 200 once at least one downstream has been polled successfully (ever), and non-200 before that. `/metrics` exposes Prometheus metrics: per-downstream last-success timestamp, poll attempt and failure counters, poll duration, emitted router count, and a merged-output generation counter. Structured logging at a configurable level: one line per downstream per cycle at debug, one line at info whenever the merged output changes, one at warning per poll failure and per rule collision.

## Acceptance criteria

- [ ] `/healthz` returns 200 even when every configured downstream is unreachable
- [ ] `/readyz` returns non-200 before any successful poll and 200 after the first successful poll of any downstream
- [ ] `/metrics` exposes last-success timestamp, attempt/failure counters, poll duration, router count, and a generation counter that increments only when merged output changes
- [ ] A log line at info appears exactly when merged output changes across two `Refresh` cycles, and not when it is unchanged
- [ ] A poll failure and a rule collision each produce a warning-level log line

gringolito:
## Summary

Adds an observability surface to `App.Handler`: `/healthz`, `/readyz`, and `/metrics`, plus DEBUG/INFO structured log lines around `Refresh`'s poll and merge cycle.

- `/healthz` always returns 200 while the process is serving, independent of downstream health.
- `/readyz` returns 503 until any downstream has ever been polled successfully, then 200 for the rest of the process lifetime.
- `/metrics` exposes Prometheus collectors on a dedicated per-App registry (not the global default, so multiple `App` instances in tests don't collide): per-downstream last-success timestamp, poll attempt and failure counters, poll duration, served router count, and a merged-output generation counter that only increments when the marshaled output actually changes.
- `Refresh` emits one DEBUG line per downstream per cycle and one INFO line exactly when the merged output changes. The existing WARN lines for poll failures and rule collisions are unchanged, now covered by tests.

## Why

Operators need a way to tell the process is alive (`/healthz`), whether it has usable data yet (`/readyz`), and what it's doing under the hood (`/metrics`, structured logs) without grepping unstructured output.

Closes #7


internal/app/app.go
```diff
@@ -205,11 +285,22 @@ func (a *App) Refresh(ctx context.Context) error {
 		}
 
 		var raw *rawdataResponse
+		// Touching the failure counter creates the per-downstream child so a
+		// zero value is still exposed before the first failure.
+		failures := a.metrics.failures.WithLabelValues(ds.Name)
+		start := a.now()
 		if r, err := a.fetchWithRetry(ctx, ds); err != nil {
```
gringolito: **Divergent Change / long method (judgement call, Fowler smell baseline)**: this poll/retry branch is where `Refresh` starts interleaving several orthogonal concerns — fetch outcome, WARN/DEBUG logging, last-known-good fallback, and (below) Prometheus counters/histogram observation, staleness withdrawal, merge, and generation counting — all in one function body. Not a hard violation (no documented repo standard on function size), but worth watching: as observability requirements keep growing this method will keep absorbing scattered edits. Consider extracting poll execution (attempt/failure/duration/last-success telemetry) into a small helper (e.g. `pollDownstream`) that `Refresh` calls once per downstream, separating domain merge logic from instrumentation.

internal/app/app.go
```diff
@@ -263,10 +355,20 @@ func (a *App) Refresh(ctx context.Context) error {
 	}
 
 	sum := sha256.Sum256(data)
+	prev := a.snap.Load()
+	if prev == nil || !bytes.Equal(prev.data, data) {
```
gringolito, line 330: Low-confidence note, not a blocker: `prev == nil` is true on the very first successful `Refresh`, so the generation counter increments and the INFO 'merged configuration changed' line fires on cycle 1 — before any "change across two Refresh cycles" as literally worded in issue #7's acceptance criteria. This reads as intentional (first publish is a change from "nothing served" to "something served", and `TestHandler_Metrics_AfterFirstRefresh`/`TestRefresh_LogsInfoExactlyOnOutputChange` both assert generation=1 / one INFO line after the first cycle), but flagging so the author can confirm this is the intended semantics of "increments only when merged output changes" rather than a literal two-cycle diff requirement.

gringolito (commented):
## Code review: Standards + Spec (issue #7)

Reviewed `main...feat/observability-endpoints` (commit 29a0e19) against the Fowler smell baseline (repo has no CODING_STANDARDS.md/CONTRIBUTING.md; `.golangci.yaml` is tooling-enforced and skipped) and against the acceptance criteria in #7.

### Standards
No documented-standard or tooling-catchable issues. One judgement-call smell: **Divergent Change / long method** — `Refresh` (internal/app/app.go) now interleaves polling, retry/backoff, staleness withdrawal, merge, output-change hashing, generation counting, and Prometheus instrumentation in one function. Left as an inline comment on the poll/metrics/log block. Not a blocker; the rest of the diff (the `metrics` type, `newMetrics()`, the new `/healthz` `/readyz` `/metrics` routes) is idiomatic Go/Prometheus-client usage and follows the file's existing conventions. Tests in `app_test.go`/`observability_test.go` are consistent with existing helper conventions (`testConfig`, `queryEndpoint`, `newSwitchableDS`, etc.).

### Spec (issue #7)
All acceptance criteria are implemented and covered by tests:
- `/healthz` always 200 regardless of downstream health — `TestHandler_Healthz_AlwaysOK`, `TestHandler_Healthz_WithAllDownstreamsDown`.
- `/readyz` non-200 before first success, 200 (cumulative) after — `TestHandler_Readyz_*`.
- `/metrics` exposes last-success gauge, attempt/failure counters, poll-duration histogram, router gauge, generation counter — `TestHandler_Metrics_*`.
- DEBUG per-downstream-per-cycle, INFO on merged-output change, WARN on poll failure and rule collision — all covered.

One low-confidence note (left inline, not a hard defect): the generation counter/INFO log also fires on the very first successful `Refresh` (prev snapshot is nil), which is a defensible reading of "output changed" (from nothing to something) and matches what the new tests lock in — flagging only so the author can confirm it's the intended semantics of "increments only when merged output changes."

No scope creep identified — the diff stays within what #7 asks for (plus the `go.mod`/`go.sum` additions needed for the Prometheus client, which are tooling output, not hand-written scope creep).

**Summary:** Standards — 1 finding (judgement call, non-blocking). Spec — 1 informational note (non-blocking, likely intentional). No P0/P1 issues found.

internal/app/app.go
```diff
@@ -191,6 +265,12 @@ func newMergeResult() *mergeResult {
 //
 // Duplicate routing rules across downstreams are logged at WARN but still
 // served, letting Traefik resolve the conflict by priority.
+//
+// Observability: one DEBUG line per downstream per cycle records the poll
+// outcome; an INFO line is emitted exactly when the merged output changes
+// across cycles (never when it is unchanged).  Poll attempts, failures,
+// duration, last-success time, and the served router count are exported as
+// Prometheus metrics on /metrics.
```
gringolito: Cut this noisy prose

internal/app/app.go
```diff
@@ -263,10 +355,20 @@ func (a *App) Refresh(ctx context.Context) error {
 	}
 
 	sum := sha256.Sum256(data)
+	prev := a.snap.Load()
+	if prev == nil || !bytes.Equal(prev.data, data) {
+		// The merged output genuinely changed: publish a new generation and
+		// log it.  Unchanged output still stores (the atomic swap for readers
+		// is preserved) but does not count as a new generation.
```
gringolito: Noisy prose, cut it

internal/app/app.go
```diff
@@ -205,11 +285,22 @@ func (a *App) Refresh(ctx context.Context) error {
 		}
 
 		var raw *rawdataResponse
+		// Touching the failure counter creates the per-downstream child so a
+		// zero value is still exposed before the first failure.
```
gringolito: Cut this noisy prose

internal/app/app.go
gringolito: Extracted pollDownstream(ctx, ds, st) in 1cdae41: it now owns attempt/failure/duration telemetry, WARN/DEBUG logging, everSucceeded/lastSuccess, stale-recovery INFO, and the last-known-good fallback. Refresh keeps only iteration, staleness withdrawal, merge, and generation/snapshot logic. Pure refactor, no behavior change — full existing test suite passes unchanged.

internal/app/app.go
gringolito: Confirmed intentional: "changed" means changed from the App's own snapshot history, and prev == nil on cycle 1 counts as a change (nothing served -> something served). TestHandler_Metrics_AfterFirstRefresh and TestRefresh_LogsInfoExactlyOnOutputChange lock in generation=1 / one INFO line after the first successful cycle. No code change made.

internal/app/app.go
gringolito: Trimmed in 1cdae41: the paragraph is now a single terse line about the exported metrics, matching the rest of the doc comment's style.

internal/app/app.go
gringolito: Trimmed in 1cdae41 to one line: "Publish a new generation only when the merged output changed."

internal/app/app.go
gringolito: Trimmed in 1cdae41 to one line: "Touching the failure counter pre-creates the per-downstream child." (now inside the extracted pollDownstream helper).

PR #21, opened 2026-09-18
feat(main): ticker-owned polling loop, signal handling, graceful shutdown

### LINKED ISSUE #8: main: ticker-owned polling loop, signal handling, graceful shutdown
## What to build

The `main` package: reads the config file path from a flag or environment variable, constructs an `App`, owns the poll ticker (calling `Refresh` on the configured interval — no time-dependent behaviour lives inside `App` itself), serves `Handler` over HTTP on the configured listen address, and shuts down cleanly on SIGTERM/SIGINT (in-flight requests finish, no dangling connections). Runnable as a plain binary with just a config file path — no other required arguments — so a problem can be reproduced locally without containers.

## Acceptance criteria

- [ ] Running the binary with a valid config file against real (or `httptest`) downstreams serves the merged config, `/healthz`, `/readyz`, and `/metrics` on the configured address
- [ ] Sending SIGTERM stops the process without an abrupt connection drop
- [ ] The poll ticker lives in `main`, not in `App` — `App`'s public surface has no timers or sleeps
- [ ] Running with a path to an invalid config file exits non-zero with the validation error from ticket 1, without starting the server

gringolito:
## Summary

Adds `cmd/traefik-scout/main.go` and `internal/cli`, wiring `config.Load`, `app.New`, and `App.Handler` into a runnable binary. `main` owns process signal handling (SIGTERM/SIGINT via `signal.NotifyContext`); `internal/cli` owns the poll ticker, HTTP server lifecycle, and graceful shutdown, so `App`'s public surface stays free of timers.

The config path resolves from the `-config` flag or the `CONFIG_PATH` environment variable, flag first. A load or validation failure prints the error to stderr and exits 1 before any server or `App` is constructed. On shutdown the ticker stops and `http.Server.Shutdown` drains in-flight requests before the process exits 0.

## Why

Issue #8 requires the binary to be runnable with just a config file path, serve the merged config over HTTP, shut down cleanly on SIGTERM/SIGINT, and keep all timing out of `App`. Splitting the wiring into `internal/cli.Run` (called by `main`) makes it testable end to end against `httptest` downstreams without touching process exit or real signals: tests assert on served `/config`, `/healthz`, `/readyz`, and `/metrics`, plus clean shutdown while a request is in flight.

Closes #8


internal/cli/cli.go
```diff
@@ -0,0 +1,141 @@
...
+	theApp, err := app.New(*cfg)
+	if err != nil {
+		_, _ = fmt.Fprintf(stderr, "traefik-scout: %v\n", err)
+		return 1
+	}
+
+	// Downstreams may be down at boot; the App serves an empty snapshot until
+	// a poll succeeds, so log and continue rather than failing startup.
+	if err := theApp.Refresh(ctx); err != nil {
+		_, _ = fmt.Fprintf(stderr, "traefik-scout: initial refresh failed: %v\n", err)
```
gringolito: The comment says this branch exists so a down-at-boot downstream is logged rather than failing startup, but `App.Refresh` never actually returns an error for a per-downstream failure — it retries, falls back to the last-known-good snapshot, logs via `a.log.WarnContext` internally, and returns `nil` unless nothing at all was contributed (which itself still returns `nil`, only `json.Marshal` failure returns non-nil). So this `if err != nil` block — and the identical pattern in `pollLoop` at line 136-137 — is effectively dead code for the scenario the comment describes; a downstream being down at boot will never print "initial refresh failed" here. Worth either removing the branch (since `Refresh` already logs failures structurally) or fixing the comment to describe what actually triggers it (a marshal error).

internal/cli/cli.go
```diff
@@ -0,0 +1,141 @@
...
+	if onListen != nil {
+		onListen(listener)
+	}
+
+	if handler == nil {
+		handler = theApp.Handler()
+	}
+	server := &http.Server{
+		Handler:           handler,
+		ReadHeaderTimeout: shutdownTimeout,
```
gringolito: `shutdownTimeout` (10s, meant to bound how long graceful shutdown waits for in-flight requests) is reused verbatim as `ReadHeaderTimeout` here — an unrelated per-connection header-read guard. The two concerns are coincidentally the same duration today, but sharing the constant couples them: tightening the shutdown grace period would silently also tighten the slow-loris guard (and vice versa). A dedicated `readHeaderTimeout` constant would make the intent at each call site unambiguous.

internal/cli/cli.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	go func() { _ = server.Serve(listener) }()
```
gringolito: `server.Serve(listener)`'s return value is discarded entirely. `Shutdown`/`Close` causing `ErrServerClosed` is the expected exit path, but any other `Serve` error (e.g. an unexpected socket failure) is silently swallowed — the process would keep running with no HTTP server and no log line, discoverable only indirectly via failed requests.

internal/cli/cli_test.go
```diff
@@ -0,0 +1,520 @@
...
+	defer cancel()
+	var stderr strings.Builder
+	done := make(chan int, 1)
+	go func() {
+		done <- cli.Run(ctx, []string{"-" + cli.FlagConfigName, path}, &stderr)
+	}()
+
+	var addr string
+	waitFor(t, 5*time.Second, func() bool {
+		for line := range strings.SplitSeq(stderr.String(), "\n") {
```
gringolito, line 295: This goroutine (via `cli.Run`) writes to `&stderr` — a `strings.Builder` — from inside `serve`/`pollLoop` (the "listening on" message here, and potentially periodic "refresh failed" messages later), while the test's main goroutine concurrently calls `stderr.String()` in the `waitFor` polling loop below with no synchronization. `strings.Builder` is documented as unsafe for concurrent use, so this is a genuine data race (would be flagged by `go test -race`, though the `mise` `test` task doesn't currently pass `-race`). Other tests in this file avoid the same trap by writing to `io.Discard` (safe for concurrent use) instead of a shared `strings.Builder`; this test should do the same, or synchronize access with a mutex.

gringolito (commented):
## Code review — Standards & Spec axes (fixed point: merge-base with `main`, `06f42b1`)

Both axes checked against issue #8's acceptance criteria (issue #2 used as background only, per scope). Overall the PR meets all four acceptance criteria: it serves the merged config/`/healthz`/`/readyz`/`/metrics`, shuts down gracefully on SIGTERM without dropping in-flight connections, keeps the ticker in `main` (App's public surface stays timer-free), and exits non-zero on invalid config before serving. Findings below are non-blocking; posted as `COMMENT`, not a merge gate.

### Standards

- **Duplicated `"traefik-scout: "` prefix** — the literal appears in seven separate `fmt.Fprintf(stderr, "traefik-scout: …")` calls across `cli.go` (`Run`, `serve`, `pollLoop`). None of the individual format strings repeat often enough to trip `goconst`, but the program-name prefix is hand-duplicated everywhere a message is printed; renaming the binary or changing the prefix format touches every call site. A small `logf(stderr, format, args...)` helper (or a `log.New(stderr, "traefik-scout: ", 0)`) would collapse this to one place. Judgement call (baseline "Duplicated Code" smell), not a hard violation.

### Spec / correctness

See inline comments for the two specific spots. Summary: the "log initial/periodic refresh failure" branches in `cli.go` never actually fire the way their comments imply (`App.Refresh` swallows per-downstream failures internally rather than returning them), and one test performs an unsynchronized concurrent read/write on a `strings.Builder`, which is a genuine data race even though it likely won't be caught today since the `test` task doesn't pass `-race`.

internal/cli/cli.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	cfgPath := *cfgFlag
+	if cfgPath == "" {
+		cfgPath = os.Getenv(EnvConfigPath)
+	}
+	if cfgPath == "" {
+		_, _ = fmt.Fprintf(stderr, "traefik-scout: no config file: use -config or set %s\n", EnvConfigPath)
+		return 1
+	}
```
gringolito, line 118: Readability: extract this over to an argument parsing method that return the parsed args, handle usage and load environment variables...

internal/cli/export_test.go
```diff
@@ -0,0 +1,22 @@
...
+// the serving lifecycle with an ephemeral listener and capture its address.
+func ServeForTest(ctx context.Context, cfg *config.Config, stderr io.Writer, onListen func(net.Listener)) int {
+	return serve(ctx, cfg, stderr, onListen, nil)
+}
+
+// ServeHandlerForTest is ServeForTest with a handler override, letting tests
+// gate requests to observe shutdown drain behavior.
+func ServeHandlerForTest(ctx context.Context, cfg *config.Config, stderr io.Writer, handler http.Handler, onListen func(net.Listener)) int {
+	return serve(ctx, cfg, stderr, onListen, handler)
+}
```
gringolito: Don't! Please don't.
I have already asked you on the past to avoid exporting just for testing, this is a hacky solution not a well designed one.

internal/cli/cli.go
gringolito: Fixed. Confirmed against internal/app/app.go's Refresh: it retries per-downstream failures, falls back to last-known-good, and logs internally, returning non-nil only on a JSON marshal failure. Rewrote both this comment and the identical one in pollLoop to state that real trigger instead of the downstream-outage scenario Refresh never actually surfaces here. Kept the log-and-continue behavior — a marshal failure is real and rare enough to still not be worth failing startup over. (commit 064d2e8)

internal/cli/cli.go
gringolito: Fixed. Added a dedicated readHeaderTimeout constant (5s) with its own doc comment, decoupled from shutdownTimeout, and used it for ReadHeaderTimeout. (commit 064d2e8)

internal/cli/cli.go
gringolito: Fixed. server.Serve()'s error is now captured and logged (via the new logf helper) when it isn't errors.Is(err, http.ErrServerClosed). (commit 064d2e8, prefix-centralization in 5377dd8)

internal/cli/cli_test.go
gringolito: Fixed. Replaced the shared strings.Builder in TestRun_ValidConfig_ServesAndShutsDownCleanly with a small mutex-guarded safeStderr type (io.Writer + String(), guarded by a sync.Mutex), used for both the writer passed to Run and the reads inside waitFor. go test -race ./internal/cli/... is clean. (commit 064d2e8)

internal/cli/cli.go
gringolito: Fixed. Extracted the flag-set construction, usage text, -config flag, fs.Parse, and CONFIG_PATH fallback out of Run into resolveConfigPath(args, stderr) (path string, code int, ok bool). Run now just calls it and returns code early on !ok; same exit codes and stderr messages as before. (commit 064d2e8)

internal/cli/export_test.go
gringolito: Understood, and you're right — my first pass at this replaced the export_test.go shim with a package-boundary workaround (flipping cli_test.go to white-box + a lint suppression), which was the same problem in different clothes. Redone properly this time, following the same precedent as WithClock/WithSleep in internal/app (commit 3184377): serve's two test-only parameters (onListen, handler) are now real exported functional options on Run — WithOnListen and WithHandler, documented, mirroring app.Option. cli_test.go is back to package cli_test like every other package's tests. The three tests that called serve directly now write a temp config file and drive everything through the public Run entry point. No exports-for-testing, no lint suppressions. (commit 064d2e8)

gringolito:
Addressed every finding from this review and the two follow-up comments:

- **Duplicated `traefik-scout: ` prefix** (this review's Standards note) — fixed. Added a `logf(stderr, format, args...)` helper collapsing all 8 call sites to one place. (commit 5377dd8)
- **Misleading Refresh-error comments**, **shutdownTimeout reused for ReadHeaderTimeout**, **discarded Serve() error**, **data race in TestRun_ValidConfig_ServesAndShutsDownCleanly**, **extract argument parsing from Run**, **remove the export_test.go side door** — all fixed, replies posted on each corresponding inline thread. (commit 064d2e8; the export_test.go fix took two passes — see the reply on that thread for what changed between them)

All fixes verified locally (`go build`, `go vet`, `go test -race ./internal/cli/...`, `mise run test`, `mise run lint`, `mise run format`) and in CI (lint, test, CodeQL all green on 5377dd8).

PR #56, opened 2026-09-18
fix(config): reject invalid timings, paths, and entrypoints at startup

### LINKED ISSUE #25: config: reject invalid timings, paths, and entrypoint lists at startup
## What to build

A configuration file that would make the service crash or silently publish nothing must be rejected at startup with a message naming the offending field, per spec #2's "configuration validated at startup with clear errors".

Verified failures on current `main`:

- `poll_interval: 0s` or a negative value starts, binds the listener, then panics with `non-positive interval for NewTicker`.
- `config_path: /metrics` (also `/healthz`, `/readyz`) starts, binds, then panics on a `ServeMux` pattern conflict.
- `edge_entrypoints` defaults to an empty list and is never checked. Emitted routers then carry no `entryPoints` key, so Traefik binds them to every entrypoint including plain HTTP. Contradicts spec user story 18.
- `max_response_size: 0` starts cleanly and then fails every poll, because the response reader is limited to zero bytes.
- `log_level: bananas` is accepted.

Also settle the `allowed_entrypoints` default. Today an empty list means "publish every router from this downstream", which inverts user story 13 (host-local routes must never reach the edge). Require at least one entrypoint per downstream so the safe posture is the only posture.

## Acceptance criteria

- [ ] Non-positive `poll_interval`, `request_timeout`, and `max_response_size` exit non-zero before the listener binds, each naming its field
- [ ] `config_path` must start with `/` and must not be `/healthz`, `/readyz`, or `/metrics`; a collision is a startup error, not a panic
- [ ] At least one `edge_entrypoints` value is required
- [ ] `log_level` accepts only the supported level names
- [ ] Each downstream requires at least one `allowed_entrypoints` value; the "empty means allow all" behaviour is removed along with the test that pinned it
- [ ] No configuration file reaches `app.Handler()` or the poll ticker in a state that can panic

gringolito:
## Summary

Extends `internal/config` startup validation to reject configurations that previously either panicked after the listener bound or silently produced broken behavior:

- `poll_interval`, `request_timeout`, and `max_response_size` must be positive.
- `config_path` must start with `/` and must not collide with the reserved `/healthz`, `/readyz`, or `/metrics` paths.
- `edge_entrypoints` requires at least one value.
- `log_level` accepts only `debug`, `info`, `warn`, or `error`.
- Each downstream's `allowed_entrypoints` now requires at least one value; the old "empty means allow all" behavior is removed, along with the test that pinned it.

Every check runs inside `config.Load`/`validate`, ahead of `app.New`, the HTTP listener bind, and the poll ticker, so no invalid configuration can reach a running service.

## Why

An empty `allowed_entrypoints` used to mean "publish every router from this downstream," which is the opposite of the opt-in allow-list the field exists for, and the other gaps let a bad config either crash mid-startup (`NewTicker` panic, `ServeMux` pattern conflict) or start cleanly while quietly misbehaving (every poll failing, or routers exposed on every entrypoint including plain HTTP). Validating these up front turns a confusing runtime failure into a startup error naming the offending field.

Closes #25


internal/config/config.go
```diff
@@ -186,6 +196,28 @@ func parseOptionalDuration(field, s string) (time.Duration, error) {
...
+	}
+	if cfg.MaxResponseSize <= 0 {
+		return fmt.Errorf("max_response_size must be positive, got %d", cfg.MaxResponseSize)
+	}
+	if !strings.HasPrefix(cfg.ConfigPath, "/") {
+		return fmt.Errorf("config_path %q must be an absolute path starting with /", cfg.ConfigPath)
+	}
+	switch cfg.ConfigPath {
+	case "/healthz", "/readyz", "/metrics":
+		return fmt.Errorf("config_path %q collides with a reserved endpoint (healthz, readyz, metrics)", cfg.ConfigPath)
```
gringolito: Judgement call (Duplicated Code / Primitive Obsession, P3): `/healthz`, `/readyz`, `/metrics` are hardcoded here, and `internal/app/app.go`'s `Handler()` hardcodes the same three strings again when registering the mux (no shared constant between the two packages). If a reserved path is renamed or added in `Handler()` without a matching edit here, this validation silently drifts out of sync and the `ServeMux` collision panic this PR sets out to prevent becomes reachable again. `goconst` won't catch this since the literals live in different packages. Worth a shared constant both `validate()` and `Handler()` reference.

gringolito (commented):
Ran the `code-review` skill against `main...HEAD` (1 commit: `8895018 fix(config): reject invalid timings, paths, and entrypoints at startup`). Spec source: issue #25, cross-checked against parent spec issue #2 (config schema section, user stories 13 & 18). Standards source: `.golangci.yaml` (no `CODING_STANDARDS.md`/`CONTRIBUTING.md` in this repo) plus the fixed Fowler smell baseline.

## Standards

No hard violations of `.golangci.yaml`-enforced rules or documented conventions were found; the new `validate()` checks follow the existing style in the same function (sequential `if`/`fmt.Errorf` guards naming the offending field, consistent with `requireAbsoluteURL` and the pre-existing duplicate-name/auth/tls checks).

One judgement-call smell (Duplicated Code / Primitive Obsession):

- `internal/config/config.go` (new `switch cfg.ConfigPath { case "/healthz", "/readyz", "/metrics": ... }`) hardcodes the same three reserved-path literals that `internal/app/app.go`'s `Handler()` already hardcodes when registering `/healthz`, `/readyz`, and `/metrics` on the mux. There is no shared constant between the two packages, so the two lists can silently drift — if a reserved path is renamed or a new one is added in `Handler()` without a matching edit in `validate()`, the exact ServeMux collision panic this PR sets out to prevent at startup becomes reachable again. `goconst` doesn't catch this because the literals live in different packages. Not a hard violation of anything documented, but worth a shared constant (e.g. in `config` or a small shared package) that both `validate()` and `Handler()` reference.

## Spec

Issue #25's acceptance criteria all appear satisfied and cleanly scoped — no missing/partial requirements, no scope creep, no implementation that looks wrong:

- Non-positive `poll_interval`/`request_timeout`/`max_response_size` now fail `config.Load` before `cli.Run` binds the listener, each naming its field (`config.go` validate()).
- `config_path` must start with `/` and colliding with `/healthz`, `/readyz`, `/metrics` is now a startup error rather than the `ServeMux` panic issue #25 describes.
- `edge_entrypoints` requires at least one value, matching user story 18 ("generated routers bound to the entrypoints I nominate at the edge") — previously an empty list silently produced routers with no `entryPoints` key.
- `log_level` is now restricted to the four `slog`-backed names.
- Each downstream now requires at least one `allowed_entrypoints` value, matching user story 13; the diff removes `TestRefreshAndHandler_EmptyAllowListPassesAll`, which is exactly the test the AC calls out as pinning the old "empty means allow all" behaviour, and `app.go`'s `mergeDownstream`/`entrypointSet` filter is simplified accordingly (the `allow != nil` bypass and the `len(eps)==0 → nil` short-circuit are both removed, consistent with the invariant validation now guarantees).
- All existing app/cli tests and testdata fixtures were updated to supply the newly-required `edge_entrypoints`/`allowed_entrypoints` fields, so no test relies on the removed permissive behaviour.

One point worth flagging for awareness rather than as a defect: `edge_entrypoints` defaults to an empty slice (`rawDefaults()`), and after this PR an empty `edge_entrypoints` is now always a hard startup error. That's exactly what issue #25 and user story 18 ask for, but it does mean `edge_entrypoints` (a global setting, not a downstream name/address) no longer has a *usable* default, which is in tension with spec #2's "every value except downstream names and addresses has a default." Since issue #25 explicitly settles this in favor of safety, this looks intentional rather than a bug — just noting the tension with the parent spec's general rule for visibility.

## Summary

- **Standards**: 1 finding, judgement call (P3) — no hard violations.
- **Spec**: 0 findings — all issue #25 acceptance criteria implemented as specified, with one pre-existing spec tension noted for awareness only.

internal/config/config.go
gringolito: Approved, we should do this.

gringolito: Addressed in a45874a: exported `config.PathHealthz`, `config.PathReadyz`, `config.PathMetrics`, and `config.ReservedPaths` from the config package. `validate()` now checks `config_path` against `ReservedPaths` instead of a hardcoded switch, and `app.Handler()` registers the mux routes using the same constants instead of its own string literals. The two can no longer drift out of sync.

PR #57, opened 2026-09-18
fix(app): skip routers whose status is not enabled

### LINKED ISSUE #27: app: skip downstream routers whose status is not enabled
## What to build

Spec #2 says to skip a router unless its status is enabled. The transform skips only the literal status `disabled`, so a router with status `warning` is republished at the edge. Verified against a fake downstream: a router with `status: "warning"` was emitted as `primary-warn` in the served document.

A warning-status router on the downstream means the downstream itself could not fully resolve it, so republishing it at the edge produces a route that points at a service the downstream will not serve.

## Acceptance criteria

- [ ] Only routers with status `enabled` are emitted
- [ ] A skipped router is logged at debug with its name and status
- [ ] A test fixture containing enabled, disabled, and warning routers asserts only the enabled one is served

gringolito:
## Summary
Only emit downstream routers whose status is exactly `enabled`, and log skipped routers at debug with their name and status.

## Why
The router filter only skipped the literal status `disabled`, so a router in any other non-enabled state (e.g. `warning`) was still republished at the edge. A warning-status router means the downstream itself could not fully resolve it, so republishing it points the edge at a route the downstream won't serve.

Closes #27


internal/app/app.go
```diff
@@ -444,7 +444,9 @@ func (a *App) mergeDownstream(ds config.Downstream, raw *rawdataResponse, result
 		if isInternalProvider(name, r.Provider) {
 			continue
 		}
-		if r.Status == "disabled" {
+		if r.Status != "enabled" {
+			a.log.Debug("skipping router with non-enabled status",
```
gringolito: Every other log call in this file that runs inside `Refresh`'s per-downstream loop uses the `*Context` variant (`DebugContext`, `WarnContext`, `InfoContext`) so cancellation/tracing context propagates with the log record (see the `WarnContext`/`DebugContext` calls a few lines above in `pollDownstream`, and `WarnContext` for stale/duplicate-rule logging in `Refresh`). This new skip log uses plain `a.log.Debug`, which drops `ctx` and is the only logger call in the file that does. Since `mergeDownstream` doesn't currently take a `ctx` parameter, either thread `ctx` through from `Refresh` (mirroring the `ds`/`raw`/`result` parameters) and call `DebugContext(ctx, ...)`, or note explicitly why this call is exempt from the pattern.

gringolito (commented):
Reviewed against issue #27's acceptance criteria and this repo's conventions (public-surface tests via `httptest.Server`, `slog`-based structured logging).

**Spec compliance:** all three acceptance criteria are met — only `status == "enabled"` routers are emitted, the skip is logged at debug with `router` and `status` fields, and `TestRefreshAndHandler_ExcludesNonEnabledStatuses` exercises enabled/disabled/warning fixtures through the public `App`/`Handler()` surface exactly like the rest of the suite.

**Standards:** one small consistency nit inline. No other issues — test style, table-driven fixtures, and helper reuse (`fakeDownstream`, `queryHandler`, `testConfig`) all match the existing patterns in `app_test.go`.

Approving; the inline note is non-blocking.

gringolito:
Isn't a metric valid here? Will it add significant value?

internal/app/app.go
gringolito: Fixed in 8fd361a: threaded `ctx` through `mergeDownstream` (mirroring `ds`/`raw`/`result`) and switched the skip log to `DebugContext(ctx, ...)`, matching every other logger call in this file's per-downstream path.

gringolito:
Good question — filed #59 to evaluate a per-downstream, per-reason skip counter separately rather than fold it into this fix. Keeping this PR scoped to #27's acceptance criteria (correct filtering + debug log); the metric decision needs its own cardinality/value tradeoff discussion, tracked in #59.

PR #58, opened 2026-09-18
feat(release): semver tag aliases, latest, and edge builds

### LINKED ISSUE #24: ci: semver tag aliases, latest, and edge builds on main
## What to build

Extend the release workflow so that a tagged release publishes full semver tag aliases plus a moving `latest`, and every push to `main` publishes a moving `edge` build — without ever letting `edge` touch `latest` or the version tags.

- Tagging release `vX.Y.Z` pushes multi-arch (`amd64`, `arm64`) images tagged `vX.Y.Z`, `vX.Y`, `vX`, and `latest`, all pointing at the same digest.
- `vX.Y` and `vX` always reflect the newest release in that series — publishing an older patch release (e.g. a backport tag) must not retroactively move a newer major/minor tag backward.
- Every push to `main` builds and pushes a multi-arch image tagged `edge`, separate from and never overwriting `latest` or any version tag.
- Opening a PR still does not publish any image (unchanged from #9).
- Image continues to run as non-root with no shell (unchanged from #9).

## Acceptance criteria

- [ ] Tagging `vX.Y.Z` pushes `amd64`+`arm64` images under `vX.Y.Z`, `vX.Y`, `vX`, and `latest`, all at the same digest
- [ ] An out-of-order older-series release does not move a newer `vX`/`vX.Y` tag backward
- [ ] A push to `main` publishes a multi-arch `edge` tag and leaves `latest` and version tags untouched
- [ ] Opening a PR publishes no image
- [ ] Published images remain non-root with no shell

gringolito:
## Summary

A tagged release `vX.Y.Z` now also publishes `vX.Y`, `vX`, and `latest` on the same multi-arch digest, and every push to `main` publishes a moving `edge` build. A new `internal/imagetags` package (used by `cmd/imagetags` inside the release workflow) decides which aliases to publish by comparing the release being tagged against the tags already published on Docker Hub, so an out-of-order backport can never move a newer `vX`/`vX.Y` alias or `latest` backward.

## Why

Issue #9 deliberately shipped without `latest` or a `main` build; its own follow-up comment names #24 as the issue that adds them once the base release pipeline existed. This closes that gap: `latest`/`vX`/`vX.Y` always track the newest release by semver rather than by tag chronology, and `edge` is owned exclusively by the CI `publish-edge` job (`push` to `main` only, never on `pull_request`) so it can never collide with the release workflow's tags.

## Reviewer notes

- `internal/imagetags/imagetags.go` is the core decision logic; `imagetags_test.go` covers the newest-release, backport, latest-follows-semver, and validation cases end to end through the package's one public function.
- `.github/workflows/release.yaml`'s "Resolve image tags" step pages the unauthenticated Docker Hub v2 API for existing tags, tolerating a first-page 404 as "no image published yet" and failing hard on any other error, then feeds the result through `cmd/imagetags` into a single `docker/build-push-action` call so every alias lands on the same digest.
- `.github/workflows/ci.yaml`'s new `publish-edge` job mirrors the existing publish steps (same pinned action SHAs) but only pushes the static `edge` tag.

Closes #24


.github/workflows/release.yaml
```diff
@@ -42,13 +47,45 @@ jobs:
...
+      # One buildx build with every resolved tag: all tags land on the same
+      # multi-arch digest.
       - uses: docker/build-push-action@c3c9e263c25d99ce0380d002d59b67737d91b0dc # v7.4.0
         with:
           context: .
           platforms: linux/amd64,linux/arm64
           push: true
-          tags: |
-            docker.io/gringolito/traefik-scout:${{ github.ref_name }}
+          tags: ${{ steps.tags.outputs.tags }}
```
gringolito, line 111: `cmd/imagetags` prints bare tags (`v1.2.3`, `v1.2`, `v1`, `latest` — see `internal/imagetags/imagetags.go`'s `Tags()` and the test fixtures, none of which include a registry/repository prefix). Feeding that straight into `docker/build-push-action`'s `tags:` input drops the `docker.io/gringolito/traefik-scout:` prefix that the replaced line (`tags: docker.io/gringolito/traefik-scout:${{ github.ref_name }}`) used to supply. `build-push-action` has no separate `images:`/`name:` input here, so `tags:` must be full image references; bare tags like `v1.2.3` will be pushed as (or rejected as) an invalid/wrong repository rather than `docker.io/gringolito/traefik-scout:v1.2.3`. This isn't exercised by CI since the release workflow only runs on a tag push, so it would only surface on the next real release. Needs the repo prefix reattached, e.g. by prefixing each line in the `Resolve image tags` step (`sed 's#^#docker.io/gringolito/traefik-scout:#'`) or having `cmd/imagetags`/`imagetags.Tags` take and prepend the image reference.

internal/imagetags/imagetags.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if ver.sameMajor(newVersion) && (maxSameMajor == nil || ver.atLeast(maxSameMajor)) {
+			maxSameMajor = &ver
+		}
+		if maxOverall == nil || ver.atLeast(maxOverall) {
+			maxOverall = &ver
+		}
+	}
+
+	tags := []string{newTag}
```
gringolito: `Tags` doesn't special-case a prerelease `newTag`. The release workflow triggers on `v*` (unchanged), which also matches full prerelease tags like `v2.0.0-rc.1`. `atLeast` compares major/minor/patch before ever looking at `pre`, so a prerelease outranks any already-published *release* in a lower series — e.g. `Tags("v2.0.0-rc.1", []string{"v1.9.9"})` returns `v2.0.0-rc.1, v2.0, v2, latest`, moving `latest`/`vX`/`vX.Y` onto a release-candidate build. The prerelease-precedence code in `atLeast`/`comparePrerelease` only breaks ties between two prereleases of the same core version, so it doesn't guard this case. If prerelease tags are ever pushed (or expected to be supported, given the amount of prerelease-comparison logic already here), this needs an explicit rule that a prerelease `newTag` never claims `latest`/`vX`/`vX.Y`.

gringolito (commented):
Two-axis review (Standards + Spec) per the code-review skill. `go vet`, `gofmt -l`, and a scoped `golangci-lint run` over the two new packages are all clean, and the new unit tests (newest-release, backport, latest-follows-semver, validation) read well and match the repo's table-free, cycle-commented Go test style. One blocking issue below would break every tagged release, plus one spec edge case worth a decision before merge.

**Standards**: no violations found beyond the inline note on `version.alias(n int)` — a minor Primitive Obsession/Mysterious-Name judgement call (magic `1`/`2` meaning "how many dot-separated components to keep"); not blocking, tooling doesn't flag it and it's easy to defend as-is.

**Spec** (issue #24): the tag-alias/backport-ordering/edge-isolation/PR-no-publish/non-root criteria are all implemented and covered by tests or by the unchanged `on: pull_request` gate. The one gap is the bare-tags-into-build-push-action issue below, which breaks acceptance criterion 1 ("pushes ... images under vX.Y.Z, vX.Y, vX, and latest") entirely for any tagged release, since no image ever actually gets built under those refs.

.github/workflows/ci.yaml
```diff
@@ -35,3 +35,30 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: cover.out
+
+  publish-edge:
+    # Pushes to main publish the moving edge image; pull_request runs never
+    # publish anything. edge is owned exclusively by this job: latest and
+    # the version tags are managed by the release workflow.
```
gringolito: Cut out this prose, it is obvious by reading the very next line.

.github/workflows/release.yaml
```diff
@@ -42,13 +47,45 @@ jobs:
...
-      # Publishes only the version tag derived from the git tag (e.g. tag
-      # v1.2.3 -> traefik-scout:v1.2.3). No floating `latest` tag is ever pushed.
+      # The build publishes the release tag itself plus the vX.Y, vX, and
+      # latest aliases. Each alias is included only when the tagged release is the
+      # newest in its semver series, so a backport release can never move a
+      # newer alias or latest backward. cmd/imagetags makes that decision
+      # from the release tags already published on Docker Hub; the Docker
+      # Hub v2 API is unauthenticated but paginated, so page through all of
+      # them. A 404 on the first page means the image does not exist yet
+      # (first-ever release, no published tags); any other error fails.
```
gringolito: Cut this out too

cmd/imagetags/main.go
```diff
@@ -0,0 +1,59 @@
+// Command imagetags prints the registry tags to publish for a release.
```
gringolito: Do we actually need to reinvent the wheel here? Isn't there anything out there available for us to use?

.github/workflows/release.yaml
gringolito: Confirmed and fixed in 4468ad8: cmd/imagetags stayed pure (bare tags), and the workflow now sources the image reference from a job-level `IMAGE` env var, piping cmd/imagetags' output through `sed "s#^#${IMAGE}:#"` before handing it to build-push-action. Verified locally: `printf 'v1.2.2\nv1.1.0\nv0.9.0\n' | go run ./cmd/imagetags v1.2.3 | sed 's#^#docker.io/gringolito/traefik-scout:#'` now produces `docker.io/gringolito/traefik-scout:v1.2.3`, `:v1.2`, `:v1`, `:latest`.

internal/imagetags/imagetags.go
gringolito: Fixed in 4468ad8: `Tags` now returns only the exact tag when `newTag` parses with a non-empty prerelease component, before any alias computation runs, so `Tags("v2.0.0-rc.1", []string{"v1.9.9"})` returns just `[v2.0.0-rc.1]`. Added `TestTags_PrereleasePublishesOnlyItself` red-green (failed against the old code with `[v2.0.0-rc.1 v2.0 v2 latest]`, passes now).

.github/workflows/ci.yaml
gringolito: Removed in 4468ad8. The job was restructured and the comment block dropped entirely; the if: condition on the job already says it.

.github/workflows/release.yaml
gringolito: Trimmed in 4468ad8. Replaced the 8-line block with a 4-line comment that only states what is not obvious from the code (that cmd/imagetags is what decides the aliases, and where the rule lives), and dropped the restated 404-handling explanation since the inline set +e/status check right below it already reads clearly.

cmd/imagetags/main.go
gringolito: Looked into this before touching anything further. docker/metadata-action's flavor.latest=auto and type=semver major/minor aliases are computed purely from the tag being built, not from what's already published, so they do not solve this: pushing an older-series backport after a newer release moves latest and the major/minor aliases backward. That's a known, still-open gap (docker/metadata-action#407), and every workaround documented there (including the maintainers' own suggestions) is a hand-rolled comparison step, either against a hard-coded env var that has to be bumped by hand on every release, or against git tags fetched at build time. I also checked "Update latest Docker tag" on the Marketplace; it only handles the latest tag, not vX/vX.Y, so we would still need custom logic for those.

Given that, I kept internal/imagetags rather than switching to metadata-action, but shrank the actual surface: it is a single pure function decided from the tags Docker Hub already reports as published (not git tags, which can drift from what is actually pushed), with the workflow doing image-reference prefixing rather than baking that into the package. Happy to swap the registry-tags source for git tags if you would rather avoid the Docker Hub v2 API call, but the ordering logic itself does not have an off-the-shelf replacement.

gringolito: I still think we can solve this with bash commands and keep the solution simple

```bash
          CURRENT_TAG="${{ github.ref_name }}"
          VERSION_NUM=${CURRENT_TAG#v}
          MAJOR=$(echo "$VERSION_NUM" | cut -d. -f1)
          MINOR=$(echo "$VERSION_NUM" | cut -d. -f2)
          GLOBAL_LATEST=$(git tag -l | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n1)
          MAJOR_LATEST=$(git tag -l | grep -E "^v?${MAJOR}\.[0-9]+\.[0-9]+$" | sort -V | tail -n1)
          MINOR_LATEST=$(git tag -l | grep -E "^v?${MAJOR}\.${MINOR}\.[0-9]+$" | sort -V | tail -n1)
          
          echo "Triggered Tag:       $CURRENT_TAG"
          echo "Global Latest Tag:   $GLOBAL_LATEST"
          echo "Major Series Latest: $MAJOR_LATEST"
          echo "Minor Series Latest: $MINOR_LATEST"
          
          # Set outputs based on scope comparisons
          if [ "$CURRENT_TAG" = "$GLOBAL_LATEST" ]; then
            echo "is_global_latest=true" >> $GITHUB_OUTPUT
          else
            echo "is_global_latest=false" >> $GITHUB_OUTPUT
          fi
          
          if [ "$CURRENT_TAG" = "$MAJOR_LATEST" ]; then
            echo "is_major_latest=true" >> $GITHUB_OUTPUT
          else
            echo "is_major_latest=false" >> $GITHUB_OUTPUT
          fi
          
          if [ "$CURRENT_TAG" = "$MINOR_LATEST" ]; then
            echo "is_minor_latest=true" >> $GITHUB_OUTPUT
          else
            echo "is_minor_latest=false" >> $GITHUB_OUTPUT
          fi
```

This example uses `git tag -l` you might reuse the tag fetching from the GitHub API (preferred) you have built over this.

gringolito: Agreed, went with this in 3c82fa6. Kept Docker Hub as the tag source rather than git tag -l though, for two reasons I flagged to the user before making the change: actions/checkout does a shallow single-ref checkout on a tag push, so git tag -l would only ever see the tag that triggered the run unless the checkout step also fetches full tag history; and a git tag existing doesn't mean its image was ever actually published (that gap is exactly what caused the P0 above). Docker Hub's tag list is immune to both.

The decision logic itself is now the bash/sort -V block you proposed, scoped to the Resolve image tags step; internal/imagetags and cmd/imagetags are gone. Verified locally against the same cases the removed Go tests covered (newest release, backport not newest in its own series, backport newest in its major but not overall, latest following semver over tag chronology, junk/invalid tags, prerelease publishing only itself, first-ever release with no prior tags), plus actionlint's shellcheck pass over both workflow files.

PR #60, opened 2026-09-18
Wire log_level and log_format into a shared logger

### LINKED ISSUE #26: cli: make log_level effective and add log_format
## What to build

`log_level` is parsed, defaulted, and then ignored. Nothing calls `slog.SetDefault` or passes a logger into the App, so it logs through the default handler at info level. Verified by running the binary with `log_level: debug` against a live downstream for two cycles: the only output was `INFO merged configuration changed routers=2`. The per-downstream debug line promised by user story 28 is unreachable in production, and the operator's main diagnostic tool for a missing route does not exist.

Build the logger from the loaded configuration and hand it to the App. Add `log_format` with `text` and `json` values, defaulting to `text`, so the spec's "structured logging" claim holds for log shippers.

The `cli` package also prints its own `traefik-scout: ...` messages with `fmt.Fprintf` alongside the App's `slog` output, so a running process emits two different log formats on stderr. Unify them.

## Acceptance criteria

- [ ] `log_level: debug` produces the per-downstream poll line every cycle; `log_level: error` suppresses the merged-output info line
- [ ] `log_format: json` emits one JSON object per line on stderr
- [ ] Startup, listen, and shutdown messages from `cli` go through the same logger and format as the App's
- [ ] No secret values appear in any log line

gringolito:
## Summary

`log_level` was parsed and validated at startup but never reached `slog`, so every process logged at the default handler's info level no matter what was configured. `cli` also printed its own `traefik-scout: `-prefixed messages via `fmt.Fprintf` alongside the App's `slog` output, so a running process emitted two different formats on stderr.

Builds one `slog.Logger` from `cfg.LogLevel` and a new `cfg.LogFormat` (`text` or `json`, defaulting to `text`), and hands it to both the App and every `cli` diagnostic message. The `logf` helper and its stderr prefix are gone.

## Why

- `log_level: debug` needs to actually produce the per-downstream poll line every cycle, and `log_level: error` needs to suppress the merged-output info line, per the spec's observability requirements.
- `log_format: json` gives log shippers one JSON object per line instead of parsing an ad hoc text prefix.
- Unifying `cli` and `App` on one logger means startup, listen, and shutdown messages share the same format as everything else the process logs.

Closes #26


gringolito (commented):
## Code review: Standards + Spec

**Spec (issue #26 acceptance criteria):** all four criteria are met.
- `log_level: debug` / `log_level: error` correctly gate the per-downstream `"downstream polled"` debug line and the `"merged configuration changed"` info line — this was already implemented in `internal/app` (via the pre-existing `WithLogger` option) and simply wasn't wired up before; this PR wires it via `serve()` → `app.New(*cfg, app.WithLogger(log))`. Verified with `TestRun_DebugLevelEmitsPerDownstreamPollLines` and `TestRun_ErrorLevelSuppressesInfoAndDebugLines`.
- `log_format: json` produces one JSON object per line via `newLogger`'s `slog.NewJSONHandler` branch, verified end-to-end by `TestRun_JSONFormatEmitsJSONLines` (parses every stderr line as JSON).
- `cli`'s own startup/config-load/listen/shutdown/serve/refresh messages all now go through the same `*slog.Logger` as the App — `logf` and its `fmt.Fprintf`/`"traefik-scout: "` prefix are fully removed (confirmed no remaining references), a clean single-format cutover.
- No secret values appear in log output — covered by `TestRun_LogsDoNotContainSecrets`, and manual inspection of `internal/app` log call sites confirms only downstream names/rule/attempt metadata are logged, never `Auth.Password`/`Auth.Headers`.

**Standards:** `go build ./...`, `go vet ./...`, and `gofmt -l` on the changed files are all clean; `go test ./internal/cli/... ./internal/config/...` passes. The commit is signed and signed-off per `CLAUDE.md`. Log key naming (`"error"`, `"downstream"`, `"addr"`) is consistent with the pre-existing `internal/app` slog call sites. One minor consistency nit inline below; nothing blocking.

Overall this is a clean, well-tested, and complete implementation of #26 — no blocking issues found.

internal/cli/cli.go
```diff
@@ -68,7 +87,7 @@ func resolveConfigPath(args []string, stderr io.Writer) (path string, code int,
 		cfgPath = os.Getenv(EnvConfigPath)
 	}
 	if cfgPath == "" {
-		logf(stderr, "no config file: use -config or set %s\n", EnvConfigPath)
+		log.ErrorContext(context.Background(), "no config file: use -config or set "+EnvConfigPath)
```
gringolito: Minor: `resolveConfigPath` logs with `context.Background()` here instead of threading the `ctx` that `Run` already has available (every other new log call in this file — `resolveConfigPath`'s callers, `serve`, `pollLoop` — passes the real request/process context). Functionally harmless with the standard handlers, but it breaks the otherwise-consistent context-threading pattern this PR establishes elsewhere. Consider adding a `ctx context.Context` parameter to `resolveConfigPath` and passing `ctx` from `Run` for consistency.

internal/cli/cli.go
```diff
@@ -112,58 +131,67 @@ func WithHandler(h http.Handler) Option {
...
 	cfg, err := config.Load(cfgPath)
 	if err != nil {
-		logf(stderr, "%v\n", err)
+		log.ErrorContext(ctx, "failed to load config", "error", err)
 		return 1
 	}
 
+	// The config-provided logger serves everything from here on: cli's own
+	// startup, listen, and shutdown messages and the App's diagnostics share
+	// one handler, so a process never emits two log formats.
```
gringolito: Cut this comment out, you don't need to add comments for obvious things.

internal/cli/cli.go
```diff
@@ -112,58 +131,67 @@ func WithHandler(h http.Handler) Option {
 // shutdownTimeout. opts applies optional behavior (WithOnListen, WithHandler);
 // production callers pass none.
 func Run(ctx context.Context, args []string, stderr io.Writer, opts ...Option) int {
-	cfgPath, code, ok := resolveConfigPath(args, stderr)
+	// Pre-config logger: config-driven level/format are unknowable before
+	// the file loads, so early diagnostics use the text default at info.
```
gringolito: This one too

internal/cli/cli.go
gringolito: Fixed in ebc7bfc: resolveConfigPath now takes a ctx context.Context parameter, and Run passes its own ctx through instead of context.Background().

internal/cli/cli.go
gringolito: Removed in ebc7bfc.

internal/cli/cli.go
gringolito: Removed in ebc7bfc.

PR #61, opened 2026-09-18
feat(cli): add self-probing container healthcheck

### LINKED ISSUE #28: image: self-probing container healthcheck on the distroless base
## What to build

The published image has no `HEALTHCHECK` and the distroless base has no shell, `curl`, or `wget`, so a Compose deployment cannot express container health at all and `depends_on: condition: service_healthy` is unavailable.

The fix is a probe mode in our own binary rather than a different base image. Prototyped and verified: added a `-healthcheck` branch to `main`, built the image with an exec-form `HEALTHCHECK`, ran it against an unreachable downstream, and Docker reported `healthy` with three consecutive exit-0 probes. Exec-form `HEALTHCHECK` needs no shell, so distroless is not a barrier.

Measured cost of the alternatives considered, uncompressed:

| Option | Base | Total image | Shell in image |
|---|---|---|---|
| Self-probe, current base | 6.18 MB | 23.17 MB (+1.1 KB) | no |
| `cgr.dev/chainguard/static` | 6.36 MB | ~23.4 MB | no (buys nothing here) |
| static busybox copied in | +1.24 MB binary | ~24.4 MB | yes, ~300 applets |
| `alpine:3` | 12.93 MB | ~29.9 MB | yes, plus apk |
| `cgr.dev/chainguard/wolfi-base` | 25.35 MB | ~42 MB | yes, plus apk |

The probe cost 1,105 bytes because `net/http` is already linked. Every base that provides `wget` also provides an attacker `sh`, which would regress user story 40.

Probe `/healthz`, not `/readyz`: readiness is 503 while no downstream has ever answered, so probing it would mark the container unhealthy for a remote host's outage, which is exactly what user story 27 exists to prevent. Take the address from the config file (`-healthcheck -config <path>`) so the probe cannot drift from the server, and dial `127.0.0.1` with the parsed port rather than the literal `listen` host, which is usually `0.0.0.0`.

## Acceptance criteria

- [ ] `traefik-scout -healthcheck -config <path>` exits 0 when the local server answers 200 on `/healthz` and non-zero otherwise
- [ ] The probe resolves its port from the config file's `listen` value and handles a wildcard host
- [ ] The image declares an exec-form `HEALTHCHECK` and `docker inspect` reports `healthy` for a running container
- [ ] The base image stays distroless nonroot; no shell, no package manager, and the image grows by less than 10 KB
- [ ] README notes that Kubernetes should use `httpGet` probes (`/healthz` for liveness, `/readyz` for readiness) instead

gringolito:
## Summary

Adds a `-healthcheck` flag to the `traefik-scout` binary that probes its own `/healthz` endpoint using the listen address from the same config file the server uses, and exits 0 on a 200 response or non-zero otherwise. The Dockerfile now declares an exec-form `HEALTHCHECK` calling this flag, and the README documents both the container healthcheck and the Kubernetes `httpGet` probe alternative.

## Why

The published image had no `HEALTHCHECK`, and the distroless base has no shell, `curl`, or `wget`, so a Compose deployment couldn't express container health and `depends_on: condition: service_healthy` was unavailable. A self-probe in the existing binary needs no shell and adds about 1KB to the image, well under the acceptance criteria's 10KB budget. The probe targets `/healthz` rather than `/readyz` because readiness is legitimately 503 while no downstream has ever answered, and the container must not be marked unhealthy for a remote host's outage.

Closes #28


internal/cli/cli.go
```diff
@@ -42,25 +42,29 @@ func logf(stderr io.Writer, format string, args ...any) {
...
 
 // resolveConfigPath parses command-line flags and the CONFIG_PATH environment
-// variable into the config file path. On a flag-parse error it prints usage
+// variable into the config file path, and reports whether -healthcheck mode
+// was requested. On a flag-parse error it prints usage
 // to stderr and returns ok=false with code 2; on a missing config path it
 // prints a hint and returns ok=false with code 1. ok=true carries the
 // resolved path.
-func resolveConfigPath(args []string, stderr io.Writer) (path string, code int, ok bool) {
+func resolveConfigPath(args []string, stderr io.Writer) (path string, healthcheck bool, code int, ok bool) {
```
gringolito: Judgement call (Standards axis, not a hard violation): the return tuple grew to 4 positional values (`path string, healthcheck bool, code int, ok bool`). This is a mild Data-Clump/readability smell — callers must track four positional slots by position. Consider a small named result type if this signature grows again.

internal/cli/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), probeTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		logf(stderr, "healthcheck: %v\n", err)
+		return 1
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
gringolito, line 42: Judgement call (Standards axis, not a hard violation): this uses the global `http.DefaultClient`, while the rest of the codebase's production HTTP calls (e.g. `internal/app/app.go`'s `buildClient`) construct a dedicated `*http.Client{Timeout: ...}`. The context deadline used here works correctly and is idiomatic, but it diverges from the established per-call client convention elsewhere in this repo.

gringolito (commented):
## Standards axis

No hard standards violations found (nothing the CI-enforced `.golangci.yaml` linters would miss). Two judgement-call smells worth a look, both low priority:

1. **`resolveConfigPath` return tuple grew to 4 positional values** (`path string, healthcheck bool, code int, ok bool`) — borderline Data Clump / readability smell; every call site now has to track four positional slots by position rather than name. Not a hard violation, just flagging for awareness as the signature grows again in the future.
2. **`runHealthcheck` uses the global `http.DefaultClient`**, whereas the established production convention in this codebase (`internal/app/app.go`'s `buildClient`) always constructs a dedicated `*http.Client{Timeout: ...}` per outbound call. The context-deadline approach used here is idiomatic Go and works correctly, but it's a minor divergence from how every other production HTTP client in the repo is built. Judgement call, not a violation.

## Spec axis

Checked against every acceptance criterion in issue #28:

- "`traefik-scout -healthcheck -config <path>` exits 0 when the local server answers 200 on `/healthz` and non-zero otherwise" — implemented in `internal/cli/healthcheck.go` and covered by `TestRun_Healthcheck_ZeroOnHealthyHealthz` / `TestRun_Healthcheck_NonZeroOnNon200` / `TestRun_Healthcheck_NonZeroOnConnectionFailure`. Verified: these tests pass.
- "The probe resolves its port from the config file's `listen` value and handles a wildcard host" — `loopbackAddr` strips the host and forces `127.0.0.1`, covered by `TestRun_Healthcheck_WildcardHostDialsLoopback`. Verified.
- "The image declares an exec-form `HEALTHCHECK` and `docker inspect` reports `healthy`" — exec-form `HEALTHCHECK` added to the Dockerfile; the `docker inspect healthy` half can't be verified from a static diff review (no CI job exercises it here), but the CMD arguments correctly resolve given the documented Compose mount path.
- "The base image stays distroless nonroot; no shell, no package manager, and the image grows by less than 10 KB" — base image unchanged, only the `HEALTHCHECK` instruction and one new source file were added; no shell or package manager introduced. Size delta can't be measured from the diff but nothing added pulls in a shell.
- "README notes that Kubernetes should use `httpGet` probes (`/healthz` for liveness, `/readyz` for readiness) instead" — added.

No missing/partial requirements, no scope creep beyond what's needed to declare a working `HEALTHCHECK` (the `--interval`/`--start-period`/`--retries` knobs are required parameters of the directive, not extra behavior), and no requirement that looks implemented incorrectly.

**Summary:** 2 findings, both Standards-axis judgement calls (Data-Clump-ish 4-value return, DefaultClient vs. the repo's per-call client convention); 0 Spec-axis findings. Both axes pass; nothing blocking.

internal/cli/cli.go
gringolito: This issue is bigger than just that. The function is called resolveConfigPath, but in fact it is actually parsing all cli arguments, the config path happens to be just one of them, the function name doesn't fit here anymore.

internal/cli/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/gringolito/traefik-scout/internal/config"
+)
+
+// FlagHealthcheckName is the -healthcheck flag name, used in usage text and
+// tests.
+const FlagHealthcheckName = "healthcheck"
+
+// probeTimeout bounds the healthcheck's HTTP request so a hung server cannot
+// hang the probe forever. Docker's HEALTHCHECK applies its own --timeout on
+// top; the probe must not rely solely on that.
```
gringolito: Short this prose

internal/cli/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+
+// runHealthcheck self-probes the running server's /healthz endpoint and
+// returns the process exit code: 0 on a 200 response, 1 otherwise. It dials
+// the address derived from the same config file the server itself uses, so
+// the probe cannot drift from the server, and never starts the poll loop or
+// binds a listener of its own.
+//
+// The probe targets /healthz, not /readyz: readiness is legitimately 503
+// while no downstream has ever answered, and the container must not be marked
+// unhealthy for a remote host's outage.
```
gringolito: Short this prose to a single paragraph, keep the obvious out of it.

internal/cli/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+func runHealthcheck(cfg *config.Config, stderr io.Writer) int {
+	addr, err := loopbackAddr(cfg.Listen)
+	if err != nil {
+		logf(stderr, "healthcheck: %v\n", err)
```
gringolito: Rebase this PR/branch against main, we have shipped a new logging structure, please use the newer stuff.

internal/cli/healthcheck.go
gringolito: Accepted violation, this is a simple healthcheck call and doesn't need to follow the app client convention.

internal/cli/healthcheck.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return 0
+}
+
+// loopbackAddr converts a configured listen address into a client-dialable
+// loopback address. The listen host is typically a wildcard such as 0.0.0.0
+// (or empty), which is a bind target rather than a dial target, so only the
+// port is kept and the host is forced to 127.0.0.1.
+func loopbackAddr(listen string) (string, error) {
+	_, port, err := net.SplitHostPort(listen)
```
gringolito: What happens if the client had set something different than localhost or 0.0.0.0? If it had hardcoded to the machine’s fixed IP address?

Dockerfile
```diff
@@ -20,4 +20,10 @@ COPY --from=build /out/traefik-scout /traefik-scout
 
 USER 65532:65532
 
+# Self-probe /healthz using the same config file the server reads, so the
+# probe cannot drift from the server. Exec form needs no shell, keeping the
+# distroless base shell-free. The probe's own timeout (3s) is shorter than
+# --timeout so a hung probe is reported, not left pending.
```
gringolito: Cut this off, it is obvious

Dockerfile
```diff
@@ -20,4 +20,10 @@ COPY --from=build /out/traefik-scout /traefik-scout
...
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 CMD ["/traefik-scout", "-healthcheck", "-config", "/etc/traefik-scout/config.yaml"]
```
gringolito: What happens if the client had set CONFIG_PATH to somewhere else?

README.md
```diff
@@ -110,6 +110,16 @@ certificate on that entrypoint for real TLS.
...
+The image declares an exec-form `HEALTHCHECK` that runs
+`/traefik-scout -healthcheck -config /etc/traefik-scout/config.yaml`: the probe reads the port from
+the same config file the server itself uses and requests `GET /healthz` on `127.0.0.1`, so it cannot
+drift from the server and no shell is needed in the image. It probes `/healthz`, never `/readyz`:
+readiness is legitimately 503 while no downstream has ever answered, and the container must not be
+marked unhealthy for a remote host's outage.
+
+Kubernetes does not use Docker's `HEALTHCHECK` mechanism. There, use `httpGet` probes instead:
+`/healthz` for liveness and `/readyz` for readiness.
+
```
gringolito: Shorten this prose to 1 or 2 sentences max.

README.md
```diff
@@ -122,4 +132,6 @@ go build -o traefik-scout ./cmd/traefik-scout
 ```
 
 The `-config` flag and `CONFIG_PATH` environment variable are equivalent. The flag takes precedence
-when both are set.
+when both are set. `traefik-scout -healthcheck -config config.yaml` probes the server configured by
+that same config file. It requests `GET /healthz` on `127.0.0.1:<listen port>` and exits 0 on a 200,
+non-zero otherwise, without starting a server of its own.
```
gringolito: Cut this off too

internal/cli/cli.go
gringolito: Fixed by renaming `resolveConfigPath` -> `parseArgs` (also resolves the naming half of this: the function now parses all CLI flags including `-healthcheck`, so a 4-value positional return matches what it actually does). See the reply below for the full rationale.

gringolito: Agreed, done: renamed to `parseArgs` in internal/cli/cli.go, updated its doc comment and the single call site in `Run`.

internal/cli/healthcheck.go
gringolito: No change, per your reply below.

gringolito: Acknowledged, left `http.DefaultClient` as-is.

internal/cli/healthcheck.go
gringolito: Trimmed to one sentence.

internal/cli/healthcheck.go
gringolito: Cut to a single tight paragraph: probe semantics plus the /healthz-not-/readyz rationale only.

internal/cli/healthcheck.go
gringolito: Rebased onto origin/main (2154313). `runHealthcheck` now takes the shared `*slog.Logger` and logs via `log.Error(...)`; the old `logf` helper is gone everywhere.

internal/cli/healthcheck.go
gringolito: Real bug, fixed: `loopbackAddr` now only rewrites wildcard/empty hosts ("", "0.0.0.0", "::") to 127.0.0.1. Any explicit host (e.g. a fixed `10.0.0.5`) is dialed as configured instead of being forced to loopback. Added table tests in internal/cli/healthcheck_internal_test.go plus two Run-level tests (IPv6 wildcard, explicit host) using real listeners.

Dockerfile
gringolito: Cut to one line.

Dockerfile
gringolito: Real bug, fixed: the HEALTHCHECK CMD no longer hardcodes `-config /etc/traefik-scout/config.yaml`; it's now `["/traefik-scout", "-healthcheck"]`. `parseArgs` already falls back to $CONFIG_PATH when `-config` is omitted, and exec-form CMD inherits the container's environment directly (no shell needed), so the probe now follows whatever CONFIG_PATH the operator actually sets.

README.md
gringolito: Cut to one sentence: what it probes and why not /readyz.

README.md
gringolito: Cut to one sentence.

